PR #4, opened 2020-10-28
added starter JSON schemas in appendix

jricher:
This adds a proposed JSON Schema to the appendix as a resource for developers. This schema currently covers the main request and response messages.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
 For example, an extension declaring an "array" representation for a field would need
 to explain how the array represents something akin to the non-array element that it
 is replacing. 
+
+# JSON Schema
+
+The following JSON Schemas are provided as a tool to help developers validate requests and
+responses within the GNAP protocol.
+
+Requesting access as in {{request}}:
+
+~~~
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "name": "Request",
+  "definitions": {
+    "resource": {
+      "type": [
+        "string",
+        "object"
+      ],
+      "required": [
+        "type"
+      ],
+      "properties": {
+        "type": {
+          "type": "string"
+        },
+        "actions": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "locations": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "datatypes": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "identifier": {
+          "type": "string"
+        }
+      }
+    },
+    "key": {
+      "type": [
+        "object"
+      ],
+      "required": [
+        "proof"
+      ],
+      "properties": {
+        "jwk": {
+          "type": "object"
+        }/* this should have a JWK-specific subschema*/,
+        "cert": {
+          "type": "string"
+        },
+        "cert#S256": {
+          "type": "string"
+        },
+        "proof": {
+          "type": "string",
+          "enum": [
+            "jwsd",
+            "jws",
+            "mtls",
+            "dpop",
+            "pop",
+            "httpsig"
+          ]/* can be expanded by extensions*/
+        }
+      }
+    },
+    "subject_id_types": {
+      "type": "string",
+      "enum": [
+        "iss-sub",
+        "email",
+        "account",
+        "phone-number"
+      ]/* more added by extensions*/
+    },
+    "assertion_types": {
+      "type": "string",
+      "enum": [
+        "id_token",
+        "saml2"
+      ]/* more added by extensions*/
+    },
+    "subject_id_structures": {
+      "type": "object",
+      "required": [
+        "subject_type"
+      ],
+      "properties": {
+        "subject_type": {
+          "$ref": "#/definitions/subject_id_types"
+        }
+      },
+      "allOf": [
+        {
+          "if": {
+            "properties": {
+              "subject_type": {
+                "const": "account"
+              }
+            }
+          },
+          "then": {
+            "properties": {
+              "uri": {
+                "type": "string"
+              }
+            },
+            "required": [
+              "uri"
+            ]
+          }
+        },
+        {
+          "if": {
+            "properties": {
+              "subject_type": {
+                "const": "email"
+              }
+            }
+          },
+          "then": {
+            "properties": {
+              "email": {
+                "type": "string"
+              }
+            },
+            "required": [
+              "email"
+            ]
+          }
+        },
+        {
+          "if": {
+            "properties": {
+              "subject_type": {
+                "const": "phone-number"
+              }
+            }
+          },
+          "then": {
+            "properties": {
+              "phone_number": {
+                "type": "string"
+              }
+            },
+            "required": [
+              "phone_number"
+            ]
+          }
+        },
+        {
+          "if": {
+            "properties": {
+              "subject_type": {
+                "const": "iss-sub"
+              }
+            }
+          },
+          "then": {
+            "properties": {
+              "iss": {
+                "type": "string"
+              },
+              "sub": {
+                "type": "string"
+              }
+            },
+            "required": [
+              "uri",
+              "sub"
+            ]
+          }
+        }
+      ]
+    }
+  },
+  "type": "object",
+  "properties": {
+    "resources": {
+      "type": [
+        "array",
+        "object"
+      ],
+      "items": {
+        "$ref": "#/definitions/resource"
+      },
+      "additionalProperties": {/* if it's an object for multiple tokens, each field's value is a resource object*/
+        "type": [
+          "array"
+        ],
+        "items": {
+          "$ref": "#/definitions/resource"
+        }
+      }
+    },
+    "subject": {
+      "type": "object",
+      "properties": {
+        "sub_ids": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/subject_id_types"
+          }
+        },
+        "assertions": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/assertion_types"
+          }
+        }
+      }
+    },
+    "client": {
+      "oneOf": [
+        {
+          "type": "object",
+          "properties": {
+            "instance_id": {
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        },
+        {
+          "type": [
+            "object",
+            "string"
+          ],
+          "properties": {
+            "key": {
+              "$ref": "#/definitions/key"
+            },
+            "class_id": {
+              "type": "string"
+            },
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+        "subject_type"
+      ],
+      "properties": {
+        "subject_type": {
+          "$ref": "#/definitions/subject_id_types"
+        }
+      },
+      "allOf": [
+        {
+          "if": {
```
yaronf, line 5474: OMG. I think the more idiomatic way is to define an object for each option (you can use `const` for the `subject_type`), and then your resulting object "OneOf" them.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+            "properties": {
+              "iss": {
+                "type": "string"
+              },
+              "sub": {
+                "type": "string"
+              }
+            },
+            "required": [
+              "uri",
```
yaronf, line 5548: uri -> iss

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+            "instance_id": {
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        },
+        {
+          "type": [
+            "object",
+            "string"
```
yaronf, line 5606: This string here is meant to represent a reference, right? If so I would instead point to a separate definition of a reference, both to make it clearer and to allow us to change the reference format in the future, in one place.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+                },
+                "uri": {
+                  "type": "string"
+                },
+                "logo_uri": {
+                  "type": "string"
+                }
+              }
+            },
+            "instance_id": false
```
yaronf, line 5632: What does this mean?

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
 For example, an extension declaring an "array" representation for a field would need
 to explain how the array represents something akin to the non-array element that it
 is replacing. 
+
+# JSON Schema
```
yaronf, line 5369: General comment: I suggest to **not** maintain this schema in the near future, because it would make changes to the protocol much more painful. Let's revisit it when the protocol settles.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+          "type": "boolean"
+        },
+        "app": {
+          "type": "boolean"
+        },
+        "ui_locales": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "pattern": "^[a-z]{2,}-[A-Z]{2,}$"/* will need a better pattern for language tags*/
```
yaronf, line 5702: No, I think we'll have to punt this one to the application.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+            "jws",
+            "mtls",
+            "dpop",
+            "pop",
+            "httpsig"
+          ]/* can be expanded by extensions*/
+        }
+      }
+    },
+    "subject_id_types": {
```
yaronf, line 5811: This would benefit from a common "include file".

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+            ]
+          }
+        }
+      ]
+    },
+    "access_token": {
+      "type": "object",
+      "properties": {
+        "value": {
+          "type": "string"
```
yaronf, line 5924: This guy is important enough to warrant a `format`.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+        },
+        "assertions": {
+          "type": "object",
+          "propertyNames": {
+            "$ref": "#/definitions/assertion_types"
+          }/* no constraints on values right now*/
+        },
+        "updated_at": {
+          "type": "string",
+          "pattern": "([0-9]{4})-([0-9]{2})-([0-9]{2})T([0-9]{2})\\:([0-9]{2})\\:([0-9]{2})"
```
yaronf, line 6027: Not sure this pattern is needed, you should have an ISO8601 parser anyway.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+    "error": { /* TODO: this should be mutually exclusive of the rest of the properties, probably -- but this is an open question */
+      "type": "string",
+      "enum": [
+        "user_denied",
+        "too_fast",
+        "unknown_request" /* To be extended */
+      ]
+    }
+  },
+  "not": {
```
yaronf, line 6046: What does this mean?

yaronf (commented):
Thank you, this must be a ton of work!

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
 is replacing. 
+
+# JSON Schema
+
+The following JSON Schemas are provided as a tool to help developers validate requests and
+responses within the GNAP protocol.
+
+Requesting access as in {{request}}:
+
+~~~
```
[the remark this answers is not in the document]
jricher: Thanks, I'll switch to that model if we do decide to include this.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
```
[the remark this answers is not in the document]
jricher: Agreed, and there are a handful of other portions of the protocol that would still need to be mapped.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+    "resource": {
+      "type": [
+        "string",
+        "object"
+      ],
+      "required": [
+        "type"
+      ],
+      "properties": {
+        "type": {
```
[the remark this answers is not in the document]
jricher: JSON schema is defined using JSON schema. The context of the word "type" here makes it unambiguous, so there's no need for a reserved word.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+          "type": "string"
+        },
+        "actions": {
```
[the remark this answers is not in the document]
jricher: I left `additionalProperties` on in most places here to allow for undefined extensions. A strict validator might want to reject them but I don't think we want to accidentally encourage people to have validators that throw out extensions when they should be ignored instead.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+        "proof"
+      ],
+      "properties": {
+        "jwk": {
+          "type": "object"
+        }/* this should have a JWK-specific subschema*/,
+        "cert": {
+          "type": "string"
+        },
+        "cert#S256": {
```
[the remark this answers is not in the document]
jricher: Thanks, filed a new issue for this one!

draft-ietf-gnap-core-protocol.md
jricher: This means `instance_id` is not allowed alongside the other properties.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5365,3 +5365,693 @@ justification for the data type representing the same basic kind of thing it ext
...
+              "type": [
+                "object",
+                "string"
+              ],
+              "properties": {
+                "name": {
+                  "type": "string"
+                },
+                "uri": {
+                  "type": "string"
```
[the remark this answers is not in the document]
jricher: Thanks, I didn't do a ton of specific string validation, this will be good to add.

draft-ietf-gnap-core-protocol.md
jricher: We can do that, but right now it's defined in the protocol as a string and so I left it as such instead of trying to abstract it ahead of time.

draft-ietf-gnap-core-protocol.md
jricher: It's as much value as defining a string field as a `uri` type, since an application would have a URI parser also.

draft-ietf-gnap-core-protocol.md
jricher: Fair point -- right now it's inherited the OAuth restrictions on access token values and so we'll need to revisit this.

draft-ietf-gnap-core-protocol.md
jricher: I think it should generally stay in sync, but I'd be comfortable with it lagging a bit during times of major churn.

draft-ietf-gnap-core-protocol.md
jricher: There probably are better ways to set it up, and I'd be open to improvements. This is one of my arguments against mutual-exclusivity of object types like this within the protocol. This is the form of polymorphism that is painful to implement correctly. :)

draft-ietf-gnap-core-protocol.md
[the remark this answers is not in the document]
yaronf: You are right.

draft-ietf-gnap-core-protocol.md
yaronf: I disagree. JSON Schema has an clean idiomatic way to represent this stuff without all the "if/else", we just need to use it.

draft-ietf-gnap-core-protocol.md
yaronf: Yes, but here you're using the schema definition to define a time format, maybe even a new time format.

jricher:
The schema has changed significantly since this was submitted, closing now and tracking the need for this in #166

PR #119, opened 2020-11-13
Moved editor's notes to external GitHub issues

jricher:


draft-ietf-gnap-core-protocol.md
```diff
@@ -4709,13 +4229,13 @@ for feedback and development of early versions of the XYZ protocol that fed into
 
 # IANA Considerations {#IANA}
 
-[[ TBD: There are a lot of items in the document that are expandable
-through the use of value registries. ]]
+\[\[ TBD: There are a lot of items in the document that are expandable
```
fimbault, line 4232: Should we generate an issue for the TBD items?

draft-ietf-gnap-core-protocol.md
```diff
@@ -1325,12 +1239,7 @@ MAY be a symmetric key known to the AS. The RC MUST NOT send a
 symmetric key by value in the request, as doing so would expose
 the key directly instead of proving possession of it. 
 
-[[ Editor's note: In many ways, passing an instance identifier
-is analogous to OAuth 2's "client_id" parameter {{RFC6749}}, especially when
-coupled with a confidential client's registration and authentication process. See
-{{example-oauth2}} for an example. Something like this is required to make things
-easier for client developers in the common case where the AS already knows
-the client's key, and to allow symmetric keys. ]]
+\[\[ [See issue #](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/) \]\]
```
fimbault: fixed by next commit

fimbault (approved):
I've rebuilt the html file, seems fine. I checked the diff and tested each link, they work (+ issue 46 in the other commit). Didn't check the correspondence between the text and the issue though.

There remains a few [[ TBD ]] at the end of the document (sections 12, 13, 14), which could probably also be issues.

draft-ietf-gnap-core-protocol.md
jricher: It doesn't hurt, but these are all very general TBD's that every draft spec needs to finish. We don't need to have linked issues for them, at least.

fimbault (approved):
Fine

PR #124, opened 2020-11-17
added type markers for protocol elements

jricher:
Mark JSON types for all member fields, addresses #6


fimbault (approved):
That is useful and clarifies where the variants are. We also see there aren't so many after all. Mostly key, resources, client, user, one knows pretty much where this happens.

Related: might need to explicit also the related errors (ex: when the client passes a string but hasn't previously provided an object). In a similar vein, should define how we deal with internal mistakes (wrong field/value) within "objects" and "array of strings", especially as we can't really define all the possible fields here.

PR #129, opened 2020-11-20
Make access token mandatory for continuation API calls.

### LINKED ISSUE #67: Continuation API artifact structure
§3.1 Request Continuation: Editor's note: 

The AS can use the optional "access_token" as a credential for the client to manage the grant request itself over time. This is in parallel with access token management as well as RS access in general. If the AS uses the access token, the continuation URL can be static, and potentially even the same as the initial request URL. If the AS does not use an access token here, it needs to use unique URLs in its response and bind the client's key to requests to those URLs - or potentially only allow one request per client at a time. The optionality adds a layer of complexity, but the client behavior is deterministic in all possible cases and it re-uses existing functions and structures instead of inventing something special just to talk to the AS. The optional access token represents a design compromise, but the working group can decide to either require the access token on all requests or to remove the access token functionality and require the security of the continuation requests be based on unique URLs.

jricher:
Closes #67

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1900,7 +1900,7 @@ access_token (object)
             RC's key used in the request and MUST NOT be a bearer token. As a consequence,
             the `key` field of this access token is always the boolean value `true`.
             This access token MUST NOT be usable at resources outside of the AS.
-            The RC MUST present the access  token in all requests to the continuation URI as 
+            The RC MUST present the access token in all requests to the continuation URI as 
             described in {{use-access-token}}.
             \[\[ [See issue #66](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/66) \]\]
 
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -2714,10 +2714,9 @@ Detached-JWS: ejy0...
 ~~~
 
 The AS MUST be able to tell from the RC's request which specific ongoing request
-is being accessed. Common methods for facilitating this include using a unique, unguessable URL
-for each continuation response, associating the request with the provided access
-token, or allowing only a single ongoing grant request for a given RC instance
-at a time. If the AS cannot determine a single active grant request to map the
+is being accessed, using a combination of the continuation URL,
+the provided access token, and the RC instance identified by the key signature.
+If the AS cannot determine a single active grant request to map the
 continuation request to, the AS MUST return an error.
 
 The ability to continue an already-started request allows the RC to perform several 
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1807,8 +1807,8 @@ as the HTTP entity body. Each possible field is detailed in the sections below
 
 
 continue (object)
-: Indicates that the RC can continue the request by making an
-    additional request using these parameters. {{response-continue}}
+: Indicates that the RC can continue the request by making one or
+    more continuation requests. {{response-continue}}
 
 access_token (object)
 : A single access token that the RC can use to call the RS on
@@ -1837,7 +1837,7 @@ error (object)
 : An error code indicating that something has gone wrong. {{response-error}}
 
 In this example, the AS is returning an [interaction URL](#response-interact-redirect),
-a [callback nonce](#response-interact-callback), and a [continuation handle](#response-continue).
+a [callback nonce](#response-interact-callback), and a [continuation response](#response-continue).
 
 ~~~
 {
@@ -1895,12 +1895,12 @@ wait (integer)
             handle and calling the URI.
 
 access_token (object)
-: RECOMMENDED. A unique access token for continuing the request, in the format specified
+: REQUIRED. A unique access token for continuing the request, in the format specified
             in {{response-token-single}}. This access token MUST be bound to the
-            RC's key used in the request and MUST NOT be a `bearer` token. 
+            RC's key used in the request and MUST NOT be a bearer token. As a consequence,
+            the `key` field of this access token is always the boolean value `true`.
             This access token MUST NOT be usable at resources outside of the AS.
-            If the AS includes an access token, the RC MUST present the access 
-            token in all requests to the continuation URI as 
+            The RC MUST present the access  token in all requests to the continuation URI as 
             described in {{use-access-token}}.
             \[\[ [See issue #66](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/66) \]\]
 
@@ -1922,7 +1922,7 @@ access_token (object)
 The RC can use the values of this field to continue the
 request as described in {{continue-request}}. Note that the
 RC MUST sign all continuation requests with its key as described
-in {{binding-keys}}. If the AS includes an `access_token`, the RC
+in {{binding-keys}} and
 MUST present the access token in its continuation request.
 
 This field SHOULD be returned when interaction is expected, to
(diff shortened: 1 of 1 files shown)


jricher:
This seems to follow the consensus of the working group as discussed at the meeting and on the list, changes will be merged unless there is a clear alternative proposal accepted by the working group.

draft-ietf-gnap-core-protocol.md
```diff
@@ -2688,33 +2688,45 @@ This is often part of facilitating [interaction](#user-interaction), but it coul
...
 Host: server.example.com
 Detached-JWS: ejy0...
 ~~~
 
-If the AS includes an `access_token` in the `continue`
-response in {{response-continue}}, the RC MUST include the access token the
-request as described in {{use-access-token}}. Note that the access token
-is always bound to the RC's presented key (or its most recent rotation).
+The AS MUST be able to tell from the RC's request which specific ongoing request
+is being accessed. Common methods for facilitating this include using a unique, unguessable URL
```
yaronf, line 2717: This is security critical (hence a MUST on the first line of the paragraph) and yet we are leaving this decision to the implementer. Why not pick one of these "common methods" and mandate it? Reducing the optionality is a **good thing** when it makes for more secure implementations.

draft-ietf-gnap-core-protocol.md
```diff
@@ -2754,9 +2755,8 @@ sections below.
 If the AS determines that the RC can 
 make a further continuation request, the AS MUST include a new 
 ["continue" response](#response-continue). 
-If the continuation was previously bound to an access token, the
-new `continue` response MUST include a bound access token as well, and
-this token SHOULD be a new access token. 
+The new `continue` response MUST include a bound access token as well, and
+this token SHOULD be a new access token, invalidating the previous access token.
```
yaronf, line 2759: Why SHOULD? Are there cases where there's a good reason to retain an old one? I am sure this detail is critical for any formal security proof.

fimbault: I fully agree this is a MUST. Dick had another view however.

yaronf: If we need to call WG consensus on this, fine. We cannot have SHOULDs just because we are unable to reach agreement. Quoting RFC 2119, "This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course." Which implies that *we* need to say when it may be appropriate to keep the old one.

draft-ietf-gnap-core-protocol.md
jricher: This is a question of API design and deployment. There is no optionality in the protocol here: it always includes a URL and an access token. We're allowing the AS to decide if it uses the access token alone, the URL, or a combination of the two in order to identify the ongoing request that the continuation request represents. If we do mandate something be the deciding factor, it should be the access token, but I'm not convinced that's really our business to say.

yaronf: Agreed, this is optionality at the implementation level, not the protocol. My push is to reduce the number of decisions implementers have to make, each of which could easily become a security issue.

draft-ietf-gnap-core-protocol.md
fimbault: The best would be for Dick to reexplain his use case (and other people as well if it makes sense) and then we can decide as a group.

draft-ietf-gnap-core-protocol.md
fimbault: Agreed. We might include a note for implementers if we think some guidance would help (just providing examples of deployment and their impact to make it concrete).

fimbault (approved):
I've looked at the rest (besides Yaron's comments), seems fine to me.

draft-ietf-gnap-core-protocol.md
```diff
@@ -2688,33 +2688,45 @@ This is often part of facilitating [interaction](#user-interaction), but it coul
...
-If the AS includes an `access_token` in the `continue`
-response in {{response-continue}}, the RC MUST include the access token the
-request as described in {{use-access-token}}. Note that the access token
-is always bound to the RC's presented key (or its most recent rotation).
+The AS MUST be able to tell from the RC's request which specific ongoing request
+is being accessed. Common methods for facilitating this include using a unique, unguessable URL
+for each continuation response, associating the request with the provided access
+token, or allowing only a single ongoing grant request for a given RC instance
+at a time. If the AS cannot determine a single active grant request to map the
+continuation request to, the AS MUST return an error.
```
jricher, line 2721: Proposed alternative text to make this MUST and its guidance more actionable:

```suggestion
The AS MUST be able to tell from the RC's request which specific ongoing request
is being accessed, using a combination of the continuation URL,
the provided access token, and the RC instance identified by the key signature.
If the AS cannot determine a single active grant request to map the
continuation request to, the AS MUST return an error.
```

yaronf: WFM.

jricher: OK, incorporated.

fimbault:
@dickhardt not sure i understand your question

draft-ietf-gnap-core-protocol.md
dickhardt: > This is security critical (hence a MUST on the first line of the paragraph) and yet we are leaving this decision to the implementer. Why not pick one of these "common methods" and mandate it? Reducing the optionality is a **good thing** when it makes for more secure implementations.

I agree with @yaronf's comments here on the optionality in implementation. This is NOT an access token in how it is usually used -- it is managing state for the AS to determine the request. A unique URL is the common mechanism. I also don't think the URL needs to be unique for each "continuation" -- it only needs to be unique for that grant request.

dickhardt:
The AS is choosing how to identify a request by a unique URL, an access token, or having only request at a time from a client. Lots of implementation choices and opportunities to make mistakes -- and overly complex.

fimbault:
There seems to be some agreement during the IETF109 call that we had to chose between the 2 alternatives. 

Now that leaves the question of which alternative to select. Between the 2, the editors have proposed (with explanations) to make access tokens mandatory.  I don't think that leaves room for mistake.

draft-ietf-gnap-core-protocol.md
```diff
@@ -1895,12 +1895,12 @@ wait (integer)
...
-: RECOMMENDED. A unique access token for continuing the request, in the format specified
+: REQUIRED. A unique access token for continuing the request, in the format specified
             in {{response-token-single}}. This access token MUST be bound to the
-            RC's key used in the request and MUST NOT be a `bearer` token. 
+            RC's key used in the request and MUST NOT be a bearer token. As a consequence,
+            the `key` field of this access token is always the boolean value `true`.
             This access token MUST NOT be usable at resources outside of the AS.
-            If the AS includes an access token, the RC MUST present the access 
-            token in all requests to the continuation URI as 
+            The RC MUST present the access  token in all requests to the continuation URI as
```
wparad, line 1903:
```suggestion
            The RC MUST present the access token in all requests to the continuation URI as 
```

jricher: Good catch, thanks!

draft-ietf-gnap-core-protocol.md
aaronpk: This PR doesn't change the behavior here, so we'll track whether this should be a SHOULD/MUST in a separate issue: #147

aaronpk (approved):

Issue #29, opened 2020-11-12
Terminology


fimbault:
People may misunderstand a term which has a specific meaning in the context of an authorization framework. The objective of this issue is to clarify the terms we use to avoid confusion.

In the current version of the GNAP spec (01), we have : 

* parties in the GNAP protocol perform actions under different roles; roles are defined by the actions taken and the expectations leveraged on the role by the overall protocol ; elements are acted upon by the roles throughout the process
* In section 1.2 (roles): Authorization Server (AS), Resource Client (RC, aka "client"), Resource Server (RS, aka "API"), Resource Owner (RO), Requesting Party (RQ, aka "user")
* In section 2.3: new notion of RC instance. This was added to allow a mecanism analogous to OAuth 2's "client_id" parameter, but it also relates to issues discussed within the OAuth WG (e.g. credential client, as an addition to public and confidential client types). It brings additional potential (attestation framework, display name, device posture, etc.)
* In section 1.3 (elements): Access Token, Grant, Cryptographic Key, Resource, Subject Information
* We also find definitions in the request/responses, e.g. resources, interaction, claims (see section 2), assertions (see 2.2)

a) are we fine with those names, abbreviations and definitions? 
- Define access token without using "credential"
- Subject information should be about the user, not the RO
- Grant as both a noun (what is being requested) and a verb?

note : we should avoid acronyms which are already widely used in our software domain (such as RP = relying party)

b) are there missing terms? 
- what is a refreshed credential? (step 7 in section 1.3.3)
- (maybe) device posture/attestation if that's something we develop further
- check all the terms in the request/responses

c) do we need a section dedicated to terminology (to define all the terms), or do we define at first use?
In ISO for instance, terms are defined at the beginning of the spec (nice for lookup). It would be interesting to provide sources to  pre-existing terminology (in our case, there's for instance https://canada-ca.github.io/PCTF-CCP/PCTF.html, GDPR/CCPA/etc.). We can then have notes (to explain what differs from a commonly accepted definition, in our specific context) and examples.
I would be in favor of doing the same. 

When commenting, please provide why you think the term and/or definition should be modified, and an alternative proposal. The objective is to prepare a PR which consolidate the changes related to terminology.

yaronf:
+1 to a separate Terminology section, for the most important 10-20 terms.

dickhardt:
+1 to a separate Terminology section. A PR for changing the terminology scattered across the document will be hard to follow.

[...]

PR #132, opened 2020-11-25
Changed "resource client (RC)" to "client instance"

jricher:
This changes all instances of "RC" and "resource client" to "CI", short for "client instance", to allow the protocol to be more precise about where it's talking about a specific instance of a piece of client software or the client software as a general class which could have many instances. This change seeks to capture some of the concepts from the design team's work that introduced the concept of a client instance more explicitly.

Addresses terminology changes, #29

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -915,7 +915,7 @@ This value is opaque to the CI and MAY be any
 valid JSON string, and therefore could include spaces, unicode
 characters, and properly escaped string sequences. However, in some
 situations the value is intended to be 
-seen and understood be the client software's developer. In such cases, the
+seen and understood by the client software's developer. In such cases, the
 API designer choosing any such human-readable strings SHOULD take steps
 to ensure the string values are not easily confused by a developer
 
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -102,7 +102,7 @@ from the design team regarding the options as understood.
 
 # Introduction
 
-This protocol allows a piece of software, the resource client, to request delegated
+This protocol allows a piece of software, the client instance, to request delegated
 authorization to resource servers and to request direct information. This delegation is
 facilitated by an authorization server usually on
 behalf of a resource owner. The requesting party operating the software may interact
@@ -135,30 +135,34 @@ on the role by the overall protocol.
 
 Authorization Server (AS)
 : Manages the requested delegations for the RO. 
-    The AS issues tokens and directly delegated information to the RC.
+    The AS issues tokens and directly delegated information to the CI.
     The AS is defined by its grant endpoint, a single URL that accepts a POST
     request with a JSON payload. The AS could also have other endpoints,
     including interaction endpoints and user code endpoints, and these are
     introduced to the RC as needed during the delegation process. 
 
-Resource Client (RC, aka "client")
+Client Instance (CI)
 : Requests tokens from the AS and uses tokens at the RS.
-    An instance of the RC software is identified by its key, which can
-    be known to the AS prior to the first request. The AS determines
-    which policies apply to a given RC, including what it can
-    request and on whose behalf. 
+    A client instance is identified by its unique key, which can
+    be known to the AS prior to the first request or introduced
+    to the AS as part of the protocol. The AS determines
+    which policies apply to a given CI, including what it can
+    request and on whose behalf. There could be many instances
+    of a single piece of client software, and where necessary this specification
+    differentiates between a specific instance, the CI, and the 
+    software running the instance, the client software.
 
 Resource Server (RS, aka "API")
-: Accepts tokens from the RC issued by the AS and serves delegated resources
+: Accepts tokens from the CI issued by the AS and serves delegated resources
     on behalf of the RO. There could be multiple RSs protected
-    by the AS that the RC will call.
+    by the AS that the CI will call.
 
 Resource Owner (RO)
-: Authorizes the request from the RC to the
+: Authorizes the request from the CI to the
     RS, often interactively at the AS.
 
 Requesting Party (RQ, aka "user")
-: Operates and interacts with the RC.
+: Operates and interacts with the CI.
 
 
 The design of GNAP does not assume any one deployment architecture,
@@ -170,26 +174,28 @@ not make additional requirements on its structure or setup.
 Multiple roles can be fulfilled by the same party, and a given party
 can switch roles in different instances of the protocol. For example,
 the RO and RQ in many instances are the same person, where a user is
-authorizing the RC to act on their own behalf at the RS. In this case,
+authorizing the CI to act on their own behalf at the RS. In this case,
 one party fulfills both of the RO and RQ roles, but the roles themselves
 are still defined separately from each other to allow for other
 use cases where they are fulfilled by different parties.
 
 For another example,
-in some complex scenarios, an RS receiving requests from one RC can act as
-an RC for a downstream secondary RS in order to fulfill the 
+in some complex scenarios, an RS receiving requests from one CI can act as
+a CI for a downstream secondary RS in order to fulfill the 
 original request. In this case, one piece of software is both an 
-RS and an RC from different perspectives, and it fulfills these 
+RS and a CI from different perspectives, and it fulfills these 
 roles separately as far as the overall protocol is concerned.
 
 A single role need not be deployed as a monolithic service. For example, 
-An RC could have components that are installed on the RQ's device as 
+A CI could have components that are installed on the RQ's device as 
 well as a back-end system that it communicates with. If both of these
 components participate in the delegation protocol, they are both considered
-part of the RC. 
+part of the CI. If there are several copies of the client software
+that run separately but all share the same key material, such as a
+deployed cluster, then this cluster is considered a single CI.
 
 For another example, an AS could likewise be built out of many constituent
-components in a distributed architecture. The component that the RC
+components in a distributed architecture. The component that the CI
 calls directly could be different from the component that the the
 RO interacts with to drive consent, since API calls and user interaction
 have different security considerations in many environments. Furthermore,
@@ -210,26 +216,26 @@ elements that are acted upon by the roles throughout the process.
 
 Access Token
 : A credential representing a set of access rights
-delegated to the RC. The access token is created by the AS, consumed
-and verified by the RS, and issued to and carried by the RC. The contents
-and format of the access token are opaque to the RC.
+delegated to the CI. The access token is created by the AS, consumed
+and verified by the RS, and issued to and carried by the CI. The contents
+and format of the access token are opaque to the CI.
 
 Grant
-: The process by which the RC requests and is given delegated
+: The process by which the CI requests and is given delegated
 access to the RS by the AS through the authority of the RO.
 
 Cryptographic Key
 : A cryptographic element binding a request to a
-holder of the key. Access tokens and RC instances can be associated with
+holder of the key. Access tokens and CI's can be associated with
 specific keys.
 
 Resource
-: A protected API served by the RS and accessed by the RC. Access to this
+: A protected API served by the RS and accessed by the CI. Access to this
 resource is delegated by the RO as part of the grant process.
 
 Subject Information
-: Information about the RO that is returned directly to the RC from the AS
-without the RC making a separate call to an RS. Access to this information
+: Information about the RO that is returned directly to the CI from the AS
+without the CI making a separate call to an RS. Access to this information
 is delegated by the RO as part of the grant process.
 
 \[\[ [See issue #33](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/33) \]\]
@@ -244,7 +250,7 @@ and not every step in this overview will happen in all circumstances.
 Note that a connection between roles in this process does not necessarily
 indicate that a specific protocol message is sent across the wire
 between the components fulfilling the roles in question, or that a 
-particular step is required every time. For example, for an RC interested
+particular step is required every time. For example, for a CI interested
 in only getting subject information directly, and not calling an RS,
 all steps involving the RS below do not apply. 
 
@@ -266,9 +272,9 @@ protocol flow.
             +                            +        
             +                            +         
         +--------+                       +       +------------+
-        |Resource|--------------(1)------+------>|  Resource  |
-        | Client |                       +       |   Server   |
-        |  (RC)  |       +---------------+       |    (RS)    |
+        | Client |--------------(1)------+------>|  Resource  |
+        |Instance|                       +       |   Server   |
+        |  (CI)  |       +---------------+       |    (RS)    |
         |        |--(2)->| Authorization |       |            |
         |        |<-(3)--|     Server    |       |            |
         |        |       |      (AS)     |       |            |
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -905,16 +911,16 @@ the AS.
 
 ~~~
 
-This value is opaque to the RC and MAY be any
+This value is opaque to the CI and MAY be any
 valid JSON string, and therefore could include spaces, unicode
 characters, and properly escaped string sequences. However, in some
 situations the value is intended to be 
-seen and understood be the RC developer. In such cases, the
+seen and understood be the client software's developer. In such cases, the
```
fimbault, line 918: well spoted

fimbault (approved):
This brings a more consistent text overall. The CI vs client software works well and has been changed throughout the text. 

I don't see any additional change to make right now. 

FYI, this PR has been highlighted in the suggested process to establish a new terminology (see https://mailarchive.ietf.org/arch/msg/txauth/XfeobMcrycw4TEAIOUkX0El1EWo/). I'll include in the proposal that we keep the term client instance - CI (possibly updating the associated definition). The only downside is the potential confusion with CI/CD (but usually the term is used in full, people don't tend to refer to CI only), so that should be fine. Then it will be interesting to check group consensus for the entire terminology section, including the item discussed here.

draft-ietf-gnap-core-protocol.md
yaronf: **by** the client software's developer

TomCJones:
I disagree. I expect to multi-home the client, so the term client instance would be incorrect. It won't even be the same software and the client will be distributed (micro services) so the concept of instance isn't even definable.  In particular i am distributing the signing so that it is NOT in the web deployed service for security reasons.
Also, i was just getting used to recourse client.  I liked it.
VOTE NO!

TomCJones:
Only slightly related point.  The idea of instance for a device hosted service of any sort is HIGHLY VALUABLE.  I would like to see an instance identifier for anything run on, for example, a user's smartphone.  Is it possible to tease these two concepts apart?

fimbault:
It is called client instance because it represents a specific occurrence of a client software (like a specific version or a specific build target), but this has currently nothing to do with the deployment model.

TomCJones:
what part of the client does the instance refer to?
1. the part that handles the HTTPS.
2. the part that handles the UX (HTML).
3. the part that signs the jws?
In heavy-use, secure implementations these are all separate software modules, probably running on different systems, on difference o/s and different compilers.
Again, for any large installation there is not a single instance to report an identifier.
There are cases were a single instance is identifiable as the "source" of the API. If that is what is meant, then that needs to be clarified. But note that is NOT the organization that is running the instance of the API.
This issue is very important in, as an example, US Health care where the programer of the API must attest to the code and that is most often not the org that is running the code. Both of these must be accessible in some manner.
Somehow we must get from the ID on the "instance" to the legal organization that takes responsibility. As is the case in privacy statements and consent receipts. As was mentioned many months ago, we need a grant request that can be passed from the client through a chain of intermediates to the resource owner to get access to the data. While the chain does not need to be inscope of GNAP, the validation of the grant request after passing thru a chain should be. The provenance of the user and the resource client must be testable at the end of the chain, the Resource Owner. That could very well include the guy that wrote the code and the guy that is asking for the data.

fimbault:
@TomCJones Really Client Instance (CI) is just a sort of generalized clientID (in OAuth2), only better suited to advanced needs  regarding dynamic device posture handling (but we're not discussing attestation for the API here), and with the requirement of proving its key possession. The CI therefore refers to any communication with the server (initial request and continuation), and that currently involves the different parts that you're describing (HTTPS, signature, etc.), but there's currently no way to separate those. 
I certainly understand the need for modularity on the server side, I'm not so sure why you'd need to split the client side also. 

But this gives me the feeling that we'd need to really understand your use case to be able to answer correctly.

yaronf:
I also think renaming the RC is strange. An RC is a client *instance* just like an AS is an authz server *instance* and similarly the RC is a particular *instance* with a certain software installed and a certain identity. The fact that we are talking about a specific instance is better explained once, in the terminology section.

dickhardt:
I agree with Yaron's comments that having "instance" in the name is
redundant.

I suggest we call it the client throughout the specification, and define it
as an specific instance of the client in the terminology. We need to
differentiate between the two server roles by an adjective (AS and RS) --
but there is only one client -- and it is the term used in OAuth and OIDC.

Also, RC indicates that the client is only accessing resources -- but in
GNAP the client may be requesting claims.



ᐧ

jricher:
A big part of what we're trying to address with this terminology change is that "client" and even "resource client" are too generic of a term here. In OAuth 2, "client" can mean a specific instance, or the software running an instance, or even just a conceptual entity registered at the AS that's running a suite of different software. OAuth 2 struggles with this by having "confidential client" and "public client" to talk about different security aspects, and OAuth 2.1 is further trying to fix this mess by introducing "credentialed client" into the mix.

The change to "client instance" here is to explicitly call out all of the places in the protocol where we mean exactly -- and only -- a single instance of client software differentiated by its key material. By not re-using the term "client" on its own from OAuth 2 we can avoid taking the luggage and false assumptions about it with us.

dickhardt:
I don't think there is as much confusion about what a client is as as you portray. In OAuth, the different adjectives about a client indicate the capabilities of the client. All the rest of the protocol is the same -- in OAuth 2.1, when client is used, it is all kinds of clients. I agree that we want to define the client as a specific instance of a client -- which is what is doing things in the protocol. Are there places in the draft where we are referring to something other than an instance of a client where we could not use an adjective to clarify?

jricher:
This pull request is explicit in calling out both "client instance" where an instance is meant and using other modifiers such as "client software" when something else is meant. What is the downside of being explicit about "client instance" throughout the document?

draft-ietf-gnap-core-protocol.md
jricher, line 918:
```suggestion
seen and understood by the client software's developer. In such cases, the
```

jricher: Fixed, thanks!

dickhardt:
> This pull request is explicit in calling out both "client instance" where an instance is meant and using other modifiers such as "client software" when something else is meant. What is the downside of being explicit about "client instance" throughout the document?

Per @yaronf comment -- we are not calling out an instance of the AS or an instance of the RS -- so we are being inconsistent between the different software -- and just calling it client has it be consistent with OAuth and OIDC -- otherwise CI makes a reader wonder how it is different -- which it is not.

fimbault:
The situation is not symmetrical here (compared to AS and RS), the client is the one that asks for authorization, so it makes sense to gather specific information on that related instance. 
It's like when someone in your household is asking for a loan to a bank, we don't really care which branch tells you yes or no, but the bank does care who you is asking specifically.

dickhardt:
@fimbault it is symmetric. The same AS or RS software can be deployed at different URLs.  Which instance of the AS or RS is what the client wants to know, just as the AS wants to know which instance of a client.

fimbault:
And beyond that I remember we agreed "client" wasn't that good a name (on the mailing list).

dickhardt:
> And beyond that I remember we agreed "client" wasn't that good a name (on the mailing list).

I don't think there was consensus on that -- or more importantly, that people took the time to participate in that discussion.

fimbault:
@dickhardt if it was symmetrical you wouldn't call them clients and servers, they would just be peers. 
Most often a client doesn't care which instance of a distributed server is running, as long as it can prove it is part of the server. A client instance on the contrary usually comes with individual characteristics of a more generic client software. 

Anyway it seems that the word "instance" gives rise to various interpretations, way beyond what was defined in the proposed text.

dickhardt:
I was not saying that the client and server were the same, nor was I giving an example of a distributed server.

If two different organizations can run the AS same software from the same vendor, each organization will have its own instance of the same AS software.

fimbault:
@dickhardt yes. But that's a totally different issue. In practice the AS you're describing is a multi-tenant architecture (physically or logically) but no one cares about device posture or attestation. There's really nothing in the protocol that we need to specify for that, it's standard IT.

dickhardt:
I was not describing a multi-tenant service -- I was describing 2 organizations independently running the same software on their own machines (or a cloud infrastructure). Each AS instance has a unique URL that is the starting endpoint. In the protocol, a client instance is communicating with an AS instance -- which in practice we call the client and the AS.

There is also client software and AS software -- which are different concepts than the client instance and the AS instance.

My point (and I believe @yaronf's point) was why do we need to call it a client instance, but not an AS instance.

fimbault:
Deploying your own server doesn't change what I explained. 

I made my best to explain the difference I see in the 2 situations.

dickhardt:
Which 2 situations are you referring to? Your last explanation had nothing to do with the point I was trying to make.

Note there is instance specific information for each AS -- it's URL, all the information returned in a discovery call, etc.

fimbault:
@dickhardt it has everything to do with it. What I explained as clearly as I can is that there are 2 distinct non symmetrical situations (client instance vs AS). 

I get that you don't agree with me, no problem with that. We should try to see what other people in the group think.

dickhardt:
I don't actually know what points you are trying to make -- do I don't think I'm disagreeing with you. 

This issue is about the term we use for the client. My position is that adding "instance" to client is odd as we are nothing adding "instance" to AS or RS

fimbault:
Yes. My position is that adding "instance" (or a similar term if people think instance is odd, but haven't really found a better name) to client makes sense and clarifies the protocol (following the new concept that extends what clientID does for Oauth2, and provides static or dynamic device specific information useful to the AS), while I see no valid reason to refer to a server "instance".

dickhardt:
I agree the term clientID has become problematic in OAuth 2. With the addition of dynamic client registration, a preregistered clientID represented something very different from a dynamically registered clientID. 

I think having a client class identifier and a client instance identifier resolves those issues. 

When describing the protocol, such as what the client sends the AS, and what the AS does -- it is always an instance of both the client and the AS that we are referring to.

Looking over the last commit that Justin made changing CI to client instance, it is unclear how adding "client instance" is better text than "client"

jricher:
The latest commit removes the defined term "client instance" in favor of being explicit throughout the protocol text where we're talking about a specific instance of the client software vs. the client software itself. Is there a reason we should not be this explicit in the protocol definition?

jricher:
The reason for not using "client" is that "client" already has a confused meaning in the OAuth 2 world and we are trying to avoid that same confusion here. Having taught hundreds of developers over the years, I can attest that the term and its use are confusing to people, and continue to be in the OAuth 2 world. 

There isn't a parallel between the AS/RS and client instance. An AS doesn't have an instance in the same way that a client does. Within one ecosystem, there's going to be an AS, regardless of which software it's running, but many clients. Depending on the type of client software, one bit of software might have many different independent instances. Those might be treated similarly, with the "class identifier" concept, but from the protocol's perspective they are all independent. This is why it's important to specify a client instance, but such specificity isn't needed with the AS.

dickhardt:
I don't agree that "client" has a confused meaning in OAuth 2 as far as the protocol is concerned - in both OAuth 2 and OIDC, the term is well known. 

Even if there are many clients in an ecosystem, they all use the protocol the same way.

From a client's perspective, there may be one client, and multiple ASes -- each with the same software, but different instances that have different characteristics. 

When reading the OAuth 2 or OIDC specs, changing "client" to "client instance" would not improve the understanding. I don't see how it helps here, and it is confusing on why we are talking about the client instance, but not the AS instance.

dickhardt:
FWIW: I think we have thrashed around this topic enough between the 3 of us. Time to get some other input in my opinion.

aaronpk:
The way I see it, whether something is a copy of a particular piece of software isn't the point, it's about entities acting within the system as a whole. 

From the perspective of the protocol, there is one AS, and it doesn't matter whether that's running the same software as another AS that's deployed separately, just as it doesn't matter if that AS is actually made up of a bunch of smaller microservices. That's acting in *the* role of the AS. There is only one AS instance in the spec.

On the other hand, there are always multiple instances of a client within the system. I agree that actually defining the term "Client Instance" is awkward, which is why I'm happy to see the actual definition changed to just "Client" now. However, using the term "instance" within the prose of the spec helps make it explicit that each instance of the client will have its own keys to use with the AS and RS.

As for OAuth, the suggestion was not to change every occurrence of "client" into "client instance", but there are certainly some cases where it would be clearer to say "client instance" there too, although that's a discussion for a different group.

draft-ietf-gnap-core-protocol.md
```diff
@@ -1294,22 +1300,22 @@ formats using a single proofing mechanism.
...
-If the RC has additional information to display to the RO
+If the client instance has additional information to display to the RO
 during any interactions at the AS, it MAY send that information in the
 "display" field. This field is a JSON object that declares information
 to present to the RO during any interactive sequences.
 
 
 name (string)
-: Display name of the RC software
+: Display name of the client instance or client software
```
aaronpk, line 1312: When would be appropriate to identify the instance vs the software here? I feel like that needs some clarification

draft-ietf-gnap-core-protocol.md
```diff
@@ -1728,9 +1734,9 @@ Additional interaction modes are defined in [a registry TBD](#IANA).
 
 \[\[ [See issue #61](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/61) \]\]
 
-## Declaring RC Capabilities {#request-capabilities}
+## Declaring Client Instance Capabilities {#request-capabilities}
```
aaronpk, line 1737: Wouldn't this be more accurate to leave out "instance" in this section? Capabilities are properties of the software.

aaronpk (changes requested):

TomCJones:
Justin has confused client instance with the code that has the key. In multihomed clients all of the instances have the ability to get a signature. There are server instances that come and go as the load comes and goes. The server does not prove anything about itself, only about the ability to get a signature.

This discussion is emblematic of the core issue --> what exactly is it that signed the message. Consider health care. The physician is a user, the client is (most often) an Electronic Health Record (EHR) server that the physician has access to. The physician issues an order to the EHR on behalf of the patient (the RO). The Client ID is of the EHR. The server can get the request signed by the EHR's key. The client instance is utterly irrelevant to the entire process. The server is picked from the active server pool.

aaronpk:
@TomCJones you're thinking too literally about the term "instance". It's not supposed to mean the instance in the ops sense, it's means exactly what you said in the first sentence, the code that has the key. That's all that matters from the perspective of the protocol.

TomCJones:
Then i hope that is clear in the doc. Yes - i use english words literally. I would hope that others could do that. I saves a whole lot of explaining outside of the doc to every person on the planet that was not here when this happened.

aaronpk:
Sorry, I should have said "narrowly" instead of "literally". The current text says:

> A client instance is identified by its unique key

That feels like it pretty accurately captures the intent of the use of "instance" in the spec. Please feel free to suggest alternative text for the definition if you feel like it doesn't accurately describe that a client instance is the code that has the key as you described in the previous comment.

fimbault:
In the Cambridge dictionary (UK) : instance = "a particular situation, event, or fact, especially an example of something that happens generally"

In US English (Merriam Webster) : "a step, stage, or situation viewed as part of a process or series of events"

Which is what we try to convey here. The key (and the fields that provide context on device posture etc.) corresponds to a specific step in the client life (e.g a version, a build etc) and the key itself might be rotated at some point.

So the use of literal English is just fine here. It just happens that in computing, and devops especially, many people use "instance" with the meaning of "copy".

jricher:
@TomCJones for a multi-homed client that has shared access to the key material, the entire structure would be considered the "instance" of the client. If that can be made more clear than the current text, quoted above, please make a suggestion!

TomCJones:
@fimbault said "It is called client instance because it represents a specific occurrence of a client software (like a specific version or a specific build target), but this has currently nothing to do with the deployment model."
During a client refresh both the old and new clients will existing at the same time. Sometimes for several days.

this could be related to section 2.3.1 & .2 issue 9
section 2.3.1 & .2 are not clear to me, but i am not sure that what i am reading is even the current state of the doc.

fimbault:
@TomCJones yes, there have been some changes based on your feedback (and Yaron and Dick). We're back on using "client", and "instance" is used in the text when it is necessary. Generally speaking, the markdown file is considered the official version, but there might be several concurrent updates through PRs are the same time. 

Regarding your point. During a client refresh, there might be both the old and new client software indeed. But a specific user would use only one of them normally (typically for a mobile app, an OTA replaces the old app by the new one / for webapps it's usually quite simple). Or do you see other cases ?

Proof method here refers to section 8. "The form of proof that the RC will use when presenting the key to the AS. The valid values of this field and the processing requirements for each are detailed in Section 8."

If other items are not clear, we'll need to clarify them (or possibly make some changes). Please let us know.

TomCJones:
I guess the question about other cases depends on whether GNAP decides to support sessions that live over a single browser tab instance. For example, in registration or renewal events i may be sent a email that requires me to click to agree. When I click i will get whatever and wherever the DNS sends me. Now it would be possible to add state to the email link, but now we are getting WAY TOO DEEP into the actual working of the site. The client ID needs some level of persistence that transcends the specific server, time zone, s/w, whatever, that might exist at the other end of a click. 

BTW, in another thread we discussed consent removal. That would most likely require a client ID consistence.

aaronpk (approved):

Issue #99, opened 2020-11-13
Updating and reading access tokens


jricher:
§6 Token Management: Editor's note: 

Should we allow for "update" to an access token by the client posting new information from a "request"? It seems this might make things weird since an access token is generally considered an unchanging thing, and the client could always request a new access token if they're allowed to continue the grant request post-issuance as in Section 5.3. There's also a possibility of being able to "read" a token's state with a GET, much like token introspection but using the token's/client's key instead of the RS key. But would a client need to "read" a token state after issuance? Is there a security risk to offering that functionality? Introspection is nearly always relegated to RS calls in practice since the client is focused on using the token at the RS. The client can always read the state of the grant itself, separately.

jricher:
See also discussion on "read" action at #98

fimbault:
Tokens are meant to be immutable. So update is not in the scope of what we want to do. 
Read access: we don't see any compelling reason to allow read access + might be an issue for privacy. Please clarify how the mecanism would work, especially with privacy considerations. 

Marked as need text. Without additional text, this section will be removed in a future update.

fimbault:
Closing the issue following the merge of the PR

PR #136, opened 2020-12-02
Clarify the nature of access requests, closes #12

### LINKED ISSUE #12: Cross-Product of Permissions
Yaron: it's implied but should be spelled out that the semantics of this example is the cross product of permissions: I am requesting to read and write and "dolphin" both "metadata" and "images", and that applies to both "locations" - 12 different possible actions.

Justin: OK, that can be expounded upon here, as that is the intent. Each object is meant to encompass one fully realized combination of options, which is why they’re presented in an array overall (as talked about above).

Yaron: this is critical to the protocol semantics.

jricher:
This added text clarifies the intent of an object within the `resources` request array, and the nature of combining multiple objects. Added examples should help explain the intended usage.

@yaronf please review to see if this addresses the concern you raise in #12

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -805,7 +805,7 @@ array represent rights of access that the RC is requesting in
 the access token. The requested access is the union of all elements
 within the array.
 
-The RC declares what access it wants to associated with the
+The RC declares what access it wants to associate with the
 resulting access token using objects that describe multiple
 dimensions of access. Each object contains a `type`
 property that determines the type of API that the RC is calling.
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -802,7 +802,7 @@ sections.
 When requesting an access token, the RC MUST send a
 `resources` field containing a JSON array. The elements of the JSON
 array represent rights of access that the RC is requesting in
-the access token. The requested access is the sum of all elements
+the access token. The requested access is the union of all elements
 within the array.
 
 The RC declares what access it wants to associated with the
@@ -880,14 +880,14 @@ The access requested for a given object when using these fields
 is the cross-product of all fields of the object. That is to 
 say, the object represents a request for all `action` values listed within the object
 to be used at all `locations` values listed within the object for all `datatype`
-values listed within the object. In the example above,
+values listed within the object. Assuming the request above was granted,
 the RC could assume that it
 would be able to do a `read` action against the `images` on the first server
 as well as a `delete` action on the `metadata` of the second server, or any other
-combination of these fields, using the same access token.
+combination of these fields, using the same access token. 
 
 To request a different combination of access, 
-such requesting one `action` against one `location` 
+such as requesting one `action` against one `location` 
 and a different `action` against a different `location`, the 
 RC can include multiple separate objects in the `resources` array.
 The following non-normative example uses the same fictitious `photo-api`
@@ -973,8 +973,8 @@ field. The second access request includes the `actions` and
 ~~~
 
 If this request is approved,
-the [resulting access token](#response-token-single) will include
-the sum of both of the requested types of access for both APIs, just as above.
+the [resulting access token](#response-token-single)'s access rights will be
+the union the requested types of access for each of the two APIs, just as above.
 
 ### Requesting Resources By Reference {#request-resource-reference}
 
@@ -1044,7 +1044,7 @@ as well as the reference values of `read`, `dolphin-metadata`, and `some other t
     ]
 ~~~
 
-The requested access is the sum of all elements of the array, including both objects and 
+The requested access is the union of all elements of the array, including both objects and 
 reference strings.
 
 \[\[ [See issue #36](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/36) \]\]
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -851,8 +851,8 @@ identifier (string)
     For example, a patient identifier for a medical API or
     a bank account number for a financial API.
 
-The following non-normative example is asking for two kinds of access to each of
-two different locations and two different data types for a single access token 
+The following non-normative example is asking for three kinds of access (read, write, delete) to each of
+two different locations and two different data types (metadata, images) for a single access token 
 using the fictitious `photo-api` type definition.
 
 ~~~
@@ -861,7 +861,8 @@ using the fictitious `photo-api` type definition.
             "type": "photo-api",
             "actions": [
                 "read",
-                "write"
+                "write",
+                "delete"
             ],
             "locations": [
                 "https://server.example.net/",
@@ -880,8 +881,10 @@ is the cross-product of all fields of the object. That is to
 say, the object represents a request for all `action` values listed within the object
 to be used at all `locations` values listed within the object for all `datatype`
 values listed within the object. In the example above,
-the RC is requesting both `read` and `write` access to two RS locations for
-both `metadata` and `images` within the API.
+the RC could assume that it
+would be able to do a `read` action against the `images` on the first server
+as well as a `delete` action on the `metadata` of the second server, or any other
+combination of these fields, using the same access token.
 
 To request a different combination of access, 
 such requesting one `action` against one `location` 
@@ -908,7 +911,8 @@ targeted access rights by using two discrete objects within the request.
         {
             "type": "photo-api",
             "actions": [
-                "write"
+                "write",
+                "delete"
             ],
             "locations": [
                 "https://resource.local/other"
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -851,18 +851,17 @@ identifier (string)
     For example, a patient identifier for a medical API or
     a bank account number for a financial API.
 
-The following non-normative example shows the use of both common
-and API-specific fields as part of two different access `type` 
-values. 
+The following non-normative example is asking for two kinds of access to each of
+two different locations and two different data types for a single access token 
+using the fictitious `photo-api` type definition.
```
fimbault, line 856: I think we should avoid any possible interpretation issue. 

Suggested change: 

The following non-normative example is asking for two kinds of access (read, write) to each of
two different locations and two different data types (metadata, images) for a single access token 
using the fictitious `photo-api` type definition.

Also we could vary the numbers so that it is even more intuitive (2 each time currently).

draft-ietf-gnap-core-protocol.md
```diff
@@ -872,6 +871,86 @@ values.
                 "metadata",
                 "images"
             ]
+        }
+    ]
+~~~
+
+The access requested for a given object when using these fields
```
fimbault, line 878: ok so that actually answers my previous comment

draft-ietf-gnap-core-protocol.md
```diff
@@ -872,6 +871,86 @@ values.
...
+        }
+    ]
+~~~
+
+The access requested here is for `read` access to `images` on one server
+while simultaneously requesting `write` access for `metadata` on a different
+server, but importantly without requesting `write` access to `images` on the
+first server.
+
+It is anticipated that API designers will use a combination
```
fimbault, line 928: since it is non normative and we don't really put requirements, just wondering if that's not better as part of an annex that shows some examples.

draft-ietf-gnap-core-protocol.md
```diff
@@ -872,6 +871,86 @@ values.
...
+of common fields defined in this specification as well as
+fields specific to the API itself. The following non-normative 
+example shows the use of both common and API-specific fields as 
+part of two different access `type` values.
```
fimbault, line 932: here geolocation

fimbault (commented):
It probably needs to be explicit as asked in #12 . But my feeling is that "cross-product of all fields" is not that clear, and it also makes the text quite long because we need to use examples in the core of the text. Which makes me wonder if some of it wouldn't belong to an annex.

Fyi, the way I translated it to myself while making sense of it: 
a) each individual resource contains additive rights (ex: read + write).
b) different sets of resources are also additive 
c) it is possible to define optional custom fields

draft-ietf-gnap-core-protocol.md
jricher: OK, these suggestions make sense. I'll expand the examples.

draft-ietf-gnap-core-protocol.md
jricher: OK, and there's also the `currency` from the second API that we can point out. I didn't do that before because I didn't want people looking for those field definitions, but I think I was overthinking it. I'll update to address that.

jricher:
@fimbault I'm in favor of having specific examples like this in-line but using an appendix for something more in-depth, like a fully specified request and response, in context with all the messages.

draft-ietf-gnap-core-protocol.md
```diff
@@ -803,7 +803,7 @@ When requesting an access token, the RC MUST send a
 `resources` field containing a JSON array. The elements of the JSON
 array represent rights of access that the RC is requesting in
 the access token. The requested access is the sum of all elements
-within the array. 
+within the array.
 
 The RC declares what access it wants to associated with the
```
yaronf, line 808: to associate

draft-ietf-gnap-core-protocol.md
```diff
@@ -851,9 +851,9 @@ identifier (string)
     For example, a patient identifier for a medical API or
     a bank account number for a financial API.
 
-The following non-normative example shows the use of both common
-and API-specific fields as part of two different access `type` 
-values. 
+The following non-normative example is asking for three kinds of access (read, write, delete) to each of
+two different locations and two different data types (metadata, images) for a single access token
```
yaronf, line 855: A comment to `locations`: the definition above is more vague than needed, the word "typically" is a red flag. Why not say that `locations` is a list of URIs that the RC can use to access Resource Servers?

draft-ietf-gnap-core-protocol.md
```diff
@@ -872,6 +872,94 @@ values.
...
             ]
+        }
+    ]
+~~~
+
+The access requested for a given object when using these fields 
+is the cross-product of all fields of the object. That is to 
+say, the object represents a request for all `action` values listed within the object
+to be used at all `locations` values listed within the object for all `datatype`
+values listed within the object. In the example above,
```
yaronf, line 883: In the example above, assuming the request is granted...

draft-ietf-gnap-core-protocol.md
```diff
@@ -872,6 +872,94 @@ values.
...
+the RC could assume that it
+would be able to do a `read` action against the `images` on the first server
+as well as a `delete` action on the `metadata` of the second server, or any other
+combination of these fields, using the same access token.
+
+To request a different combination of access, 
+such requesting one `action` against one `location`
```
yaronf, line 890: such as

draft-ietf-gnap-core-protocol.md
```diff
@@ -953,6 +1044,9 @@ string-type resource items.
     ]
 ~~~
 
+The requested access is the sum of all elements of the array, including both objects and
```
yaronf, line 1047: I think "union" is a better word than "sum".

draft-ietf-gnap-core-protocol.md
```diff
@@ -953,6 +1044,9 @@ string-type resource items.
     ]
 ~~~
 
+The requested access is the sum of all elements of the array, including both objects and 
+reference strings.
+
```
yaronf, line 1049: I'm not sure if this belongs here or should be a separate issue: we should be clear on the semantics of the request/response.
* If the response contains a `resources` object, the RC is granted **exactly** this access. 
* If it doesn't, the RC is granted **exactly** the access that was requested.

The word "exactly" here means that anything outside this access (other actions, other locations, or any unspecified combination) MUST be rejected by the RC.

draft-ietf-gnap-core-protocol.md
```diff
@@ -886,7 +974,7 @@ values.
 
 If this request is approved,
 the [resulting access token](#response-token-single) will include
-the sum of both of the requested types of access.
+the sum of both of the requested types of access for both APIs, just as above.
```
yaronf, line 977: I think "union" is a better word than "sum".

yaronf (changes requested):
OK in general, but see my comments.

draft-ietf-gnap-core-protocol.md
jricher: The reason I've initially held back on defining `locations` as URIs is that ultimately it's up to the RS to recognize and enforce this, so it might not be a URI that the client accesses, it could be an identifier for the RS known by the AS/RS coupling. We could make this more strict now and see if there's need in practice for loosening up this field or defining a new field, but that should be a separate issue (not the topic of this PR).

draft-ietf-gnap-core-protocol.md
jricher: I think this is a separate issue. I'm personally more in favor of the response always containing the `resources` or equivalent for every access token. OAuth 2 has a similar optimization with `scope` where you don't have to return it if you don't want to, but I think it might make more sense to just always return it and the client can ignore it if it doesn't need anything from it.

draft-ietf-gnap-core-protocol.md
fimbault: yes you're right

draft-ietf-gnap-core-protocol.md
fimbault: I think Yaron's question is interesting. It may well be that the request is very different from the actual access that is provided. We might indeed always return a resource section, which could potentially just return "same as requested" if there was no limitation. Or we could always return it in full. 
But in practice what matters is the access token itself, which is really the source of truth. Do we plan to implement token introspection (and is it mandatory or not)?

-> probably a separate issue.

jricher: Introspection is a service for RS's and not clients (see #99), and so an introspection response might be different from the access token response to the client. 

I've opened @yaronf 's comment in a new issue: #141.

draft-ietf-gnap-core-protocol.md
jricher: Raised as a new issue in #142.

draft-ietf-gnap-core-protocol.md
fimbault: If not called introspection, wasn't there something similar (maybe that was called management Uri if I recall correctly?)

jricher: I'm marking this conversation resolved so that it can continue in the new issue.

draft-ietf-gnap-core-protocol.md
```diff
@@ -885,8 +973,8 @@ values.
 ~~~
 
 If this request is approved,
-the [resulting access token](#response-token-single) will include
-the sum of both of the requested types of access.
+the [resulting access token](#response-token-single)'s access rights will be
+the union the requested types of access for each of the two APIs, just as above.
```
yaronf, line 977: the union of

yaronf (approved):
LGTM, except for one typo (separate comment).

draft-ietf-gnap-core-protocol.md
jricher, line 977:
```suggestion
the union of the requested types of access for each of the two APIs, just as above.
```

jricher: Good eye, fixed!

Issue #63, opened 2020-11-13
Splitting the "claims" request


jricher:
§2.8 Requesting OpenID Connect Claims: Editor's note: 

In order to use the "claims" parameter as defined in OIDC, we have to violate the principle of orthogonality in Section 2.9. An alternative approach would be to split up the portions of the claims request, so that "id_token" claims would go into the "subject" field and "userinfo" claims would go into the "resources" request, but this violates the original field definition from OIDC and gets into the territory of defining an identity schema request. This approach would also invalidate extensions to the "claims" standard as each "target" would need to have its own separate mapping to some part of the GNAP protocol.

fimbault:
This is informational, should follow the fate of #64

PR #138, opened 2020-12-03
drop examples of requesting OIDC claims

### LINKED ISSUE #64: Including OpenID Connect claims
§2.8 Requesting OpenID Connect Claims: Editor's note: 

I'm not a fan of GNAP defining how OIDC would work at all and would rather that work be done by the OIDF in an extension. However, I think it is important for discussion to see this kind of thing in context with the rest of the protocol, for now. In the future, I would anticipate this would be defined by the OIDF as a relatively small but robust identity layer on top of GNAP.

aaronpk:
closes #64

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -73,15 +73,6 @@ normative:
           ins: B. de Medeiros
         -
           ins: C. Mortimore
-    OIDC4IA:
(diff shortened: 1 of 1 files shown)


jricher (approved):

jricher:
Also closes #63

aaronpk:
Apologies, I accidentally clicked the merge button too early before we went through the documented process of adding the "pending merge" tag. I've recreated this PR as #140.

PR #139, opened 2020-12-03
drop "Redirect to a Shortened URL"

### LINKED ISSUE #53: Error mode for short interaction URL
§2.5.1.1 Redirect to an Arbitrary Shortened URL

> If the AS cannot shorten its response URL enough to fit in the requested size, the AS SHOULD return an error.

Or maybe just ignore this part of the interaction request?

### LINKED ISSUE #121: Redirect to an Arbitrary Shortened URL
I don't understand why this mode is needed?

Its also problematic having a boolean or a number as valid parameters for `interact.redirect`

aaronpk:
closes #121 and closes #53

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1496,6 +1496,7 @@ device, but it cannot accept a callback.
 
 ~~~
     "interact": {
+        "redirect": true,
         "user_code": true
     }
 ~~~
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4455,9 +4455,8 @@ Content-type: application/json
 
 In this scenario, the user does not have access to a web browser on
 the device and must use a secondary device to interact with the AS.
-The client can display a user code or a printable QR code. The client
-prefers a short URL if one is available, with a maximum of 255 characters
-in length. The is not able to accept callbacks from the AS and needs to poll
+The client can display a user code or a printable QR code.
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -1490,13 +1490,10 @@ a browser request.
 ~~~
 
 In this non-normative example, the RC is indicating that it can 
-display a [user code](#request-interact-usercode) and direct the RQ
-to an [arbitrary URL of maximum length](#request-interact-short) 255
-characters, but it cannot accept a callback.
+display a [user code](#request-interact-usercode), but it cannot accept a callback.
```
jricher, line 1493: This example can keep the redirection piece, just not the shortened form.

```suggestion
display a [user code](#request-interact-usercode) and direct the RQ
to an [arbitrary URL](#request-interact-redirect) on a secondary
device, but it cannot accept a callback.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1490,13 +1490,10 @@ a browser request.
...
 
 ~~~
     "interact": {
```
jricher, line 1496:
```suggestion
    "interact": {
        "redirect": true,
```

jricher (changes requested):
The same change needs to be made in the appendix under "Secondary Device Interaction {#example-device}". 

One of these examples can keep the redirection as an option alongside the user code, just not the shortened form.

aaronpk:
Thanks, I missed that!

jricher (approved):

Issue #134, opened 2020-12-02
Guidance for Extensions


fimbault:
regrouping #65 #61 and #80 to provide guidance

yaronf:
I would appreciate more detail here:
- A finite list of specific extension points that we want to enable (additional claims in the request? in the response?). IMO supporting too many extension points leads to code complexity, hence security issues. It also makes it harder to formally verify the protocol.
- Features that we trimmed and that we'd like to enable to reinstate if someone cares about them (maybe signature methods?).
- Additional requirements (having named extensions? allowing a protocol party to declare that one or more extensions are critical?).

We should aim for a strict protocol definition (e.g. messages can be validated using JSON Schema) where implementations refuse to accept unexpected "stuff", and only allow very specific places where additional "stuff" is acceptable. This may not be best practice for JSON-using protocols, but it is for security protocols.

yaronf:
I agree with @fimbault  in principle, but let's not go too far. I think that SAML2 is functionality that'll be widely used from day one (correct me if I am wrong). If so, it needs to stay in the core spec.

fimbault:
After discussion between editors, that's currently what will be proposed (draft8 includes the related security considerations).

Remains the question of JOSE, which will be discussed at IETF112. Personally I believe a separate draft would make sense (detached and attached JWS could be used by the oauth community as well) and it would also be easier to rally people that dislike JOSE to the GNAP WG. I short we'd be agnostic to those subjective assessments

jricher:
Editors will list all intended extension points within Core, probably pulling together all the existing stubs for extensions in different sections into a larger discussion. This would require at least a shell of the IANA section for creating necessary registries.

jricher:
Closed by #436

PR #152, opened 2020-12-23
Refactor "key" section

### LINKED ISSUE #130: Refactor key presentation
Currently, the `key` field is defined in the client's initial request but its format is referenced in a number of other places throughout the spec. It's awkward calling back to this request field, so we should factor the key definition and its semantics to its own separate section. I think it would fit well under the "binding keys" section of the document since that also defines the presentation mechanisms.

jricher:
Refactors the key, token, and proofing sections into a new presentation section. Cleans up language around key presentation and binding. 

Also cleans up a few items missed by the merge of #132 

closes #130


draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3213,99 @@ the token management URL as valid, since the end result is still the token
...
+use in that part of the protocol. 
+
+~~~
+    "key": "S-P4XJQ_RYJCRTSU1.63N3E"
+~~~
+
+Keys referenced in this manner MAY be shared keys. The key reference
+MUST NOT contain any unencrypted private or shared key information.
+
+The means of dereferencing this value are out of scope for this specification.
```
yaronf, line 3299: I disagree: when used between the Client and AS this is clearly required for interop. Feel free to open an issue.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3213,99 @@ the token management URL as valid, since the end result is still the token
...
+
+Keys in GNAP can also be passed by reference such that the party receiving
+the reference will be able to determine the appropriate keying material for
+use in that part of the protocol. 
+
+~~~
+    "key": "S-P4XJQ_RYJCRTSU1.63N3E"
+~~~
+
+Keys referenced in this manner MAY be shared keys. The key reference
```
yaronf, line 3296: "Shared key" is IMO not a common term. Suggest to say "shared symmetric keys".

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3213,99 @@ the token management URL as valid, since the end result is still the token
...
+            contain an "alg" field which is used to validate the signature.
+            MUST contain the "kid" field to identify the key in the signed
+            object.
+
+cert (string)
+: PEM serialized value of the certificate used to
+            sign the request, with optional internal whitespace.
 
-If the key value is the boolean `false`, the access token is a bearer token
+cert#S256 (string)
```
yaronf, line 3262: This is not a key type (it's not a key!) so should not be listed here. Or if it is, we need to explain why.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3334,22 +3372,21 @@ MUST validate all components of the signed message to ensure that nothing
 has been tampered with or substituted in a way that would change the nature of
 the request.
 
+When a keep proofing mechanism is bound to an access token, the access token MUST be covered
```
yaronf, line 3375: keep -> key

yaronf: Also, I think this statement is aimed at us (the protocol developers) rather than implementers. IOW, I suggest to remove it.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3213,99 @@ the token management URL as valid, since the end result is still the token
...
+In GNAP, the client instance secures its requests to the AS and RS by presenting an access
+token, presenting proof of a key that is possesses, or both an access token and
+key proof together.
+
+- When an access token is used with a key proof, this is a bound token request. This type of
+    request is used for calls to the RS as well as the AS during negotiation.
+- When a key proof is used with no access token, this is a non-authorized signed request. This
+    type of request is used for calls to the AS to initiate a negotiation.
+- When an access token is used with no key proof, this is a bearer token request. This type of
+    request is used only for calls to the RS.
```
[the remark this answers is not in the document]
jricher: I don't think it's helpful to frame this in terms of preferences, as an RS might not be the party deciding which tokens it accepts or how (that is, the AS ultimately decides which tokens are issued for RS's it protects).

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3213,99 @@ the token management URL as valid, since the end result is still the token
...
-If the key value is the boolean `false`, the access token is a bearer token
+cert#S256 (string)
+: The certificate thumbprint calculated as
+            per [OAuth-MTLS](#RFC8705) in base64 URL
+            encoding.
+
+Additional key types are defined in [a registry TBD](#IANA).
+
+This non-normative example shows a single key presented in multiple
+formats using a single proofing mechanism.
```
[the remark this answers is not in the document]
jricher: I can see how that can be misread, we can reword it. The "single proofing mechanism" refers to the value of the `proof` field, which is singular here.

draft-ietf-gnap-core-protocol.md
jricher: I disagree that we need to specify this method as it's up to the AS to know how to interpret the value and likely up to the AS to issue such a value to the client. So long as the AS knows which key the client instance means here it should be fine. Opened up a new issue for it #154

draft-ietf-gnap-core-protocol.md
jricher: This statement is aimed any anyone who would write a key proofing method to extend the set here. Flexibility in this regard is foundational to the protocol. I think the sentiment needs to stay but it can be reworded or relocated if needed (though I'd suggest a separate PR/issue for this, perhaps under #134 that's collecting these items).

draft-ietf-gnap-core-protocol.md
jricher: It's enough information for the AS/RS to validate the key presented in the TLS layer matches the one referenced here. We can expand the language on what goes in this object if that's helpful.

yaronf: Yes please. Something like: This "key type" is a reference to the key presented at the TLS layer.

draft-ietf-gnap-core-protocol.md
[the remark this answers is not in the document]
yaronf: Got it. But from a clarity POV, you still want to mention explicitly that the same use case (Client to RS) can use either of the options.

Also, some people are looking for a looser binding of the AS to RS, so you should mention that we assume tight coupling between them. Consider the case of versioning: RS switches from unbound to bound tokens. The AS must know about it and orchestrate this change.

draft-ietf-gnap-core-protocol.md
jricher: I've tried to clarify the nature of this value, and also removed the words "key type" that I think were misleading.

yaronf (approved):

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3215,106 @@ the token management URL as valid, since the end result is still the token
 not being usable.
 
 
-# Using Access Tokens {#use-access-token}
+# Securing Requests from the Client Instance {#secure-requests}
+
+In GNAP, the client instance secures its requests to the AS and RS by presenting an access
+token, presenting proof of a key that it possesses, or both an access token and
+key proof together.
```
fimbault, line 3222: both an access token and key proof together.
-> key proof of possession ?

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3215,106 @@ the token management URL as valid, since the end result is still the token
...
+
+- When an access token is used with a key proof, this is a bound token request. This type of
+    request is used for calls to the RS as well as the AS during negotiation.
+- When a key proof is used with no access token, this is a non-authorized signed request. This
+    type of request is used for calls to the AS to initiate a negotiation.
+- When an access token is used with no key proof, this is a bearer token request. This type of
```
fimbault, line 3228: here also it's a key proof of possession, it we want to be explicit about it.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3261,12 +3215,106 @@ the token management URL as valid, since the end result is still the token
...
+by value. Proof of this key material MUST be bound to a request, the nature of which varies with
+the location in the protocol the key is used. For a key used as part of a client instance's initial request
+in {{request-client}}, the key value is the client instance's public key, and
+proof of that key MUST be presented in that request. For a key used as part of an
+access token response in {{response-token-single}}, the proof of that key MUST
+be used when presenting the access token.
+
+A key presented by value MUST be a public key in at least one
+supported format. If a key is sent in multiple
+formats, all the key format values MUST be equivalent. Note that
```
fimbault, line 3247: what do you mean exactly by equivalent? (just to be sure)

draft-ietf-gnap-core-protocol.md
```diff
@@ -3276,11 +3324,11 @@ Authorization: Bearer OS9M2PMHKUR64TB8N6BW7OZB8CDFONP219RP1LT0
 The form parameter and query parameter methods of {{RFC6750}} MUST NOT
 be used.
 
-If the "key" value is the boolean `true`, the access token MUST be sent
-to the RS using the same key and proofing mechanism that the client instance used
+If the `key` value is the boolean `true`, the access token MUST be sent
```
fimbault, line 3327: maybe this should be review when we talk about key rotation, because it might not be the same as the initial

fimbault (commented):
The changes are ok, I made a few suggestions, but no big deal

draft-ietf-gnap-core-protocol.md
jricher: It means they represent the same underlying key. So an RSA key in X509 format under `cert` can also be represented as a JWK under `jwk`, and either of these could be with or without CRT values, and all would be equivalent as defined by the RSA key type. If there's a reference for cryptographic key equivalence we should cite it here.

draft-ietf-gnap-core-protocol.md
jricher: This is "key proof" below as well, all of which mean "proof of possession of a specific key". Should we add a defined term for this, perhaps? At the moment it's at least consistent if we later replace with a better term.

draft-ietf-gnap-core-protocol.md
jricher: This is a good point, I'll add in the rotation language here to be consistent with other parts of the spec.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3276,11 +3324,11 @@ Authorization: Bearer OS9M2PMHKUR64TB8N6BW7OZB8CDFONP219RP1LT0
...
+using the same key and proofing mechanism that the client instance used
 in its initial request.
```
jricher, line 3329:
```suggestion
in its initial request (or its most recent rotation).
```

draft-ietf-gnap-core-protocol.md
fimbault: ok. If we see it is useful to add a definition, we can do that later on (right now I'd keep the current scope, since there are already quite a few changes, but I put that in my notes)

draft-ietf-gnap-core-protocol.md
fimbault: ok

fimbault (approved):

PR #153, opened 2020-12-27
fix small bug about sub_ids in grant response

nerocrux:
Should be a bug in `3.4. Returning User Information`.
Please take a look. Thank you!


jricher (approved):

yaronf:
@jricher Could you fix the workflow so that PRs from outside the editorial team do not fail the PR checks? Thanks and a Happy 2021!

Issue #145, opened 2020-12-13
Opaque / Non opaque access token


fimbault:
Currently we assume that tokens are opaque to the client, meaning that the client doesn't care about their internals (although in practice JWT is often used). It's even described as such in our current version of the terminology (v02) : "The contents and format of the access token are opaque to the RC."

Yet, there has been discussion on the IETF mailing list that for privacy reasons, there might be a need for non opaque tokens, so that its content can be verified (i.e. doesn't include more rights than requested). 

Related actions : do we want to support non opaque tokens?  
1) review the potential impact on the definition of "access token" in #29 
2) should we define a datastructure for GNAP access tokens? (JSON Web Token (JWT) Profile for OAuth 2.0 Access Tokens or equivalent). How do we keep flexibility if one needs to extend the default format (more fields), or even integrate with something different than JWT? (we don't want to limit potential innovation here).
3) does it replace introspection/management endpoints?
4) should it be enforced? is it just an option?

jricher:
I am against the client having any expectation of being able to understand the access token. If a component needs to send a message directly to the client in the way the client can understand it, then it should use a different protocol element to do so.

The common trust model in the OAuth world is that the client is the untrusted component, which is why it needs to get the access token in the first place.

fimbault:
I have mixed feelings here. 
I agree with the client being unstrusted from a security perspective, what we're saying here is that the AS is untrusted from a privacy perspective. 

To me, the biggest potential downside of non opaque tokens is that it potentially limits innovation. A big part of the success of JWT comes from json's flexibility. Plus nothing stops me in OAuth from using a completely different format if I wish. That's really valuable. My current experience here would be to keep that behavior.
But it does come at the expense of not having a schema that can be verified by default, which could come handy in some cases. The alternative is introspection/management APIs, but there's always the possibility that the token could actually be different than the related resource description.  

So to go forward, I guess Denis should make its case here, and then we can take a decision. This should also balance the alternatives.

jricher:
It's also a privacy issue to release information to the client in many cases. Several profiles of OAuth that I've worked on prohibit the token itself to carry user information or identifiers in a way that the client could read. In these cases, it's the business of the RS and the AS, and the token is just the artifact to carry rights between them. 

[...]

PR #155, opened 2021-01-05
update terminology

fimbault:
This corresponds to #29 and the related [wiki](https://github.com/ietf-wg-gnap/gnap-core-protocol/wiki/Terminology), as well as advances on the mailing list. 

Sections 1.2 and 1.3 have been updated accordingly.
The term "cryptographic keys" has been removed. Some additional terms have been added (especially sub-definitions, such as "right", "attribute", "privilege").
The term RQ has been renamed to end-user and changes in the main text.

I kept a simple layout for notes and examples, maybe this can be improved.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -125,17 +125,17 @@ Roles are defined by the actions taken and the expectations leveraged
 on the role by the overall protocol. 
 
 Authorization Server (AS)
-: server that grants privileges to a particular end-user and that provides them to a client in the form of an access token. 
+: server that grants delegated privileges to a particular instance of client software in the form of an access token and other information (such as subject information). 
 
 Client
 : application operated by an end-user that consumes resources from one or several RSs, possibly requiring access privileges from one or several ASs. 
 
-    Note: this specification differentiates between a specific instance (the Client instance, identified by its public key) and the software running the instance (the Client software). For some kinds of Client software, there could be many instances of a single piece of Client software.
+    Example: a client can be a mobile application, a web application, etc.
 
-    Example: a Client can be a mobile application, a web application, etc.
+    Note: this specification differentiates between a specific instance (the client instance, identified by its unique key) and the software running the instance (the client software). For some kinds of client software, there could be many instances of that software, each instance with a different key.
 
 Resource Server (RS)
-: server that provides operations on resources where protected operations require that the client provides one or more valid access tokens issued by one or more ASs.
+: server that provides operations on protected resources, where operations require a valid access token issued by an AS. 
 
 Resource Owner (RO)
 : subject entity that may grant or deny operations on resources it has authority upon.
@@ -143,7 +143,7 @@ Resource Owner (RO)
     Note: the act of granting or denying an operation may be manual (i.e. through an interaction with a physical person) or automatic (i.e. through predefined organizational rules).
 
 End-user 
-: natural person that operates with the client software.
+: natural person that operates the client instance.
 
     Note: that natural person may or may not be the same entity as the RO. When it is explicit that the end-user and the RO are the same entity, the generic term user may be used. 
 
@@ -195,36 +195,34 @@ In addition to the roles above, the protocol also involves several
 elements that are acted upon by the roles throughout the process.
 
 Attribute
-: characteristics related to an end-user.
+: characteristics related to a subject.
 
 Access Token
-: digitally signed data that contains specific rights and/or attributes.
+: a data artifact representing a set of rights and/or attributes.
 
-    Note: an access token can be first issued to an end-user (usually requiring his authentication) and subsequently refreshed.
-
-Claim
-: statement about a subject.
+    Note: an access token can be first issued to an client instance (requiring authorization by the RO) and subsequently rotated.
 
 Grant
-: (verb): to permit an end-user to exercise some rights and/or to assert some attributes at a specific time and during a specific duration. (noun): the act of granting.
+: (verb): to permit an instance of client software to exercise some set of delegated rights to access a protected resource and/or to receive some attributes at a specific time and during a specific duration. (noun): the act of granting.
 
 Privilege
-: right or attribute associated with an end-user.
+: right or attribute associated with a subject.
 
-Resource
-: public or protected resource from a RS.
+    Note: the RO defines and maintains the rights and attributes associated to the protected resource, and might temporarily delegate some set of those privileges to an end-user. This process is refered to as privilege delegation. 
 
 Protected Resource
 : protected API (Application Programming Interface) served by a RS and that can be accessed by a client, if and only if a valid access token is provided.
 
-Right
-: ability given to an end-user to perform a given operation on a resource under the control of a RS.
+    Note: to avoid complex sentences, the specification document may simply refer to resource instead of protected resource.   
 
-Subject Entity
+Right
+: ability given to a subject to perform a given operation on a resource under the control of a RS.
+
+Subject
 : person, organization or device.
 
 Subject Information
-: claim asserted locally by an AS about a subject entity.
+: statement asserted locally by an AS about a subject.
 
 ## Sequences {#sequence}
 
@@ -283,10 +281,10 @@ protocol flow.
     ~ ~ ~ indicates a potential equivalence or out-of-band communication between roles
 ~~~
 
-- (A) The End-user interacts with the client instance to indicate a need for resources on
+- (A) The end-user interacts with the client instance to indicate a need for resources on
     behalf of the RO. This could identify the RS the client instance needs to call,
     the resources needed, or the RO that is needed to approve the 
-    request. Note that the RO and End-user are often
+    request. Note that the RO and end-user are often
     the same entity in practice.
     
 - (1) The client instance [attempts to call the RS](#rs-request-without-token) to determine 
@@ -306,7 +304,7 @@ protocol flow.
     using a variety of possible mechanisms including web page
     redirects, applications, challenge/response protocols, or 
     other methods. The RO approves the request for the client instance
-    being operated by the End-user. Note that the RO and end-user are often
+    being operated by the end-user. Note that the RO and end-user are often
     the same entity in practice.
 
 - (4) The client instance [continues the grant at the AS](#continue-request).
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -96,7 +96,7 @@ from the design team regarding the options as understood.
 This protocol allows a piece of software, the client instance, to request delegated
 authorization to resource servers and to request direct information. This delegation is
 facilitated by an authorization server usually on
-behalf of a resource owner. The requesting party operating the software may interact
+behalf of a resource owner. The end-user operating the software may interact
 with the authorization server to authenticate, provide consent, and
 authorize the request.
 
@@ -125,37 +125,27 @@ Roles are defined by the actions taken and the expectations leveraged
 on the role by the overall protocol. 
 
 Authorization Server (AS)
-: Manages the requested delegations for the RO. 
-    The AS issues tokens and directly delegated information to an instance of the client.
-    The AS is defined by its grant endpoint, a single URL that accepts a POST
-    request with a JSON payload. The AS could also have other endpoints,
-    including interaction endpoints and user code endpoints, and these are
-    introduced to the RC as needed during the delegation process. 
+: server that grants privileges to a particular end-user and that provides them to a client in the form of an access token. 
 
 Client
-: Requests tokens and directly delegated information from the AS, and uses tokens at the RS.
-    For some kinds of client software, there could be many instances
-    of a single piece of client software. This specification
-    differentiates between a specific instance (the client instance) and the 
-    software running the instance (the client software).
-    A client instance is identified by its unique key, which can
-    be known to the AS prior to the first request or introduced
-    to the AS as part of the protocol. The AS determines
-    which policies apply to a given client instance, including what it can
-    request and on whose behalf. 
+: application operated by an end-user that consumes resources from one or several RSs, possibly requiring access privileges from one or several ASs. 
 
-Resource Server (RS, aka "API")
-: Accepts tokens from the client instance issued by the AS and serves delegated resources
-    on behalf of the RO. There could be multiple RSs protected
-    by the AS that the client instance will call.
+    Note: this specification differentiates between a specific instance (the Client instance, identified by its public key) and the software running the instance (the Client software). For some kinds of Client software, there could be many instances of a single piece of Client software.
+
+    Example: a Client can be a mobile application, a web application, etc.
+
+Resource Server (RS)
+: server that provides operations on resources where protected operations require that the client provides one or more valid access tokens issued by one or more ASs.
 
 Resource Owner (RO)
-: Authorizes the request from the client instance to the
-    RS, often interactively at the AS.
+: subject entity that may grant or deny operations on resources it has authority upon.
 
-Requesting Party (RQ, aka "user")
-: Operates and interacts with the client instance.
+    Note: the act of granting or denying an operation may be manual (i.e. through an interaction with a physical person) or automatic (i.e. through predefined organizational rules).
 
+End-user 
+: natural person that operates with the client software.
+
+    Note: that natural person may or may not be the same entity as the RO. When it is explicit that the end-user and the RO are the same entity, the generic term user may be used. 
 
 The design of GNAP does not assume any one deployment architecture,
 but instead attempts to define roles that can be fulfilled in a number
@@ -165,9 +155,9 @@ not make additional requirements on its structure or setup.
 
 Multiple roles can be fulfilled by the same party, and a given party
 can switch roles in different instances of the protocol. For example,
-the RO and RQ in many instances are the same person, where a user is
+the RO and end-user in many instances are the same person, where a user is
 authorizing the client instance to act on their own behalf at the RS. In this case,
-one party fulfills both of the RO and RQ roles, but the roles themselves
+one party fulfills both of the RO and end-user roles, but the roles themselves
 are still defined separately from each other to allow for other
 use cases where they are fulfilled by different parties.
 
@@ -179,7 +169,7 @@ RS and a client instance from different perspectives, and it fulfills these
 roles separately as far as the overall protocol is concerned.
 
 A single role need not be deployed as a monolithic service. For example, 
-A client instance could have components that are installed on the RQ's device as 
+A client instance could have components that are installed on the end-user's device as 
 well as a back-end system that it communicates with. If both of these
 components participate in the delegation protocol, they are both considered
 part of the client instance. If there are several copies of the client software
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -125,37 +125,27 @@ Roles are defined by the actions taken and the expectations leveraged
...
-    on behalf of the RO. There could be multiple RSs protected
-    by the AS that the client instance will call.
+: application operated by an end-user that consumes resources from one or several RSs, possibly requiring access privileges from one or several ASs. 
+
+    Note: this specification differentiates between a specific instance (the Client instance, identified by its public key) and the software running the instance (the Client software). For some kinds of Client software, there could be many instances of a single piece of Client software.
+
+    Example: a Client can be a mobile application, a web application, etc.
+
+Resource Server (RS)
+: server that provides operations on resources where protected operations require that the client provides one or more valid access tokens issued by one or more ASs.
```
jricher, line 138: Suggest we remove "client" from this as the RS isn't concerned with (and often doesn't know about) the client software, just the token. I know there was some discussion on this but I also don't care for the "one or more" items here. By saying "a token" and "an AS" we aren't limiting the RS to also accept others, but it's defined by accepting at least one.

```suggestion
: server that provides operations on resources where protected operations require a valid access token issued by an AS trusted by the RS for this purpose.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -125,37 +125,27 @@ Roles are defined by the actions taken and the expectations leveraged
...
-: Authorizes the request from the client instance to the
-    RS, often interactively at the AS.
+: subject entity that may grant or deny operations on resources it has authority upon.
+
+    Note: the act of granting or denying an operation may be manual (i.e. through an interaction with a physical person) or automatic (i.e. through predefined organizational rules).
 
-Requesting Party (RQ, aka "user")
-: Operates and interacts with the client instance.
+End-user 
+: natural person that operates with the client software.
```
jricher, line 146: This is a case where the end user is using a specific instance of the client software for purposes of the protocol, not just someone who operates the software in general.

```suggestion
: natural person that operates the client instance.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,37 @@ role of the AS as defined by the protocol.
...
+: characteristics related to an end-user.
+
 Access Token
-: A credential representing a set of access rights
-delegated to the client instance. The access token is created by the AS, consumed
-and verified by the RS, and issued to and carried by the client instance. The contents
-and format of the access token are opaque to the client.
+: digitally signed data that contains specific rights and/or attributes.
+
+    Note: an access token can be first issued to an end-user (usually requiring his authentication) and subsequently refreshed.
```
jricher, line 203: Tokens aren't issued to people, they're issued to software -- also the spec currently uses "rotate" instead of "refresh" language, though that might not be wholly consistent yet so we should check that.

```suggestion
    Note: an access token can be first issued to an client instance (requiring authorization by the RO) and subsequently rotated.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,37 @@ role of the AS as defined by the protocol.
...
 
+Attribute
+: characteristics related to an end-user.
+
 Access Token
-: A credential representing a set of access rights
-delegated to the client instance. The access token is created by the AS, consumed
-and verified by the RS, and issued to and carried by the client instance. The contents
-and format of the access token are opaque to the client.
+: digitally signed data that contains specific rights and/or attributes.
```
jricher, line 201: A token needn't be signed and needn't contain any information in it whatsoever. Reference-based tokens have neither, for example. Suggest we keep the wider definition. 

```suggestion
: a data artifact representing a set of delegated access rights, the format of which is opaque to the client software.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,37 @@ role of the AS as defined by the protocol.
...
+
+    Note: an access token can be first issued to an end-user (usually requiring his authentication) and subsequently refreshed.
+
+Claim
+: statement about a subject.
```
jricher, line 206: I don't think this is a helpful definition, especially if we only use it in the context of "subject information" below. It's a very general term to the point of being almost meaningless.

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,37 @@ role of the AS as defined by the protocol.
...
 
 Grant
-: The process by which the client instance requests and is given delegated
-access to the RS by the AS through the authority of the RO.
+: (verb): to permit an end-user to exercise some rights and/or to assert some attributes at a specific time and during a specific duration. (noun): the act of granting.
```
jricher, line 209: The rights aren't delegated to the user, they're delegated to the client software, usually to act on behalf of the user. Also the assertion of attributes is :from: the AS, not from the end-user.

```suggestion
: (verb): to permit an instance of client software to exercise some set of delegated rights to access a protected resource and/or to receive some attributes at a specific time and during a specific duration. (noun): the act of granting such delegated rights.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,37 @@ role of the AS as defined by the protocol.
...
 
-Cryptographic Key
-: A cryptographic element binding a request to a
-holder of the key. Access tokens and client instances can be associated with
-specific keys.
+Privilege
+: right or attribute associated with an end-user.
```
jricher, line 212: Are these associated with the end user or the resource owner?

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,37 @@ role of the AS as defined by the protocol.
...
 
 Resource
-: A protected API served by the RS and accessed by the client instance. Access to this
-resource is delegated by the RO as part of the grant process.
+: public or protected resource from a RS.
+
+Protected Resource
+: protected API (Application Programming Interface) served by a RS and that can be accessed by a client, if and only if a valid access token is provided.
```
jricher, line 218: What's the value of separating "resource" and "protected resource"? Do we have an "unprotected resource" in scope?

draft-ietf-gnap-core-protocol.md
```diff
@@ -285,10 +283,10 @@ protocol flow.
     ~ ~ ~ indicates a potential equivalence or out-of-band communication between roles
 ~~~
 
-- (A) The RQ interacts with the client instance to indicate a need for resources on
+- (A) The End-user interacts with the client instance to indicate a need for resources on
```
jricher, line 286:
```suggestion
- (A) The end-user interacts with the client instance to indicate a need for resources on
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -285,10 +283,10 @@ protocol flow.
...
     behalf of the RO. This could identify the RS the client instance needs to call,
     the resources needed, or the RO that is needed to approve the 
-    request. Note that the RO and RQ are often
+    request. Note that the RO and End-user are often
```
jricher, line 289:
```suggestion
    request. Note that the RO and end-user are often
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -308,7 +306,7 @@ protocol flow.
     using a variety of possible mechanisms including web page
     redirects, applications, challenge/response protocols, or 
     other methods. The RO approves the request for the client instance
-    being operated by the RQ. Note that the RO and RQ are often
+    being operated by the End-user. Note that the RO and end-user are often
```
jricher, line 309:
```suggestion
    being operated by the end-user. Note that the RO and end-user are often
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -357,10 +355,10 @@ that returns from the interaction.
 
 ~~~
     +--------+                                  +--------+         +------+
-    | Client |                                  |   AS   |         |  RO  |
-    |Instance|                                  |        |         |  +   |
-    |        |< (1) + Start Session + + + + + + + + + + + + + + + +|  RQ  |
-    |        |                                  |        |         |(User)|
+    | Client |                                  |   AS   |         | User |
+    |Instance|                                  |        |         |      |
```
jricher, line 359:
```suggestion
    | Client |                                  |   AS   |         | End  |
    |Instance|                                  |        |         | User |
```

jricher (changes requested):
Several definitions need clean up, but a lot of this is an improvement!

draft-ietf-gnap-core-protocol.md
yaronf: I don't think we should include the "opaque" part here: this is an implementation detail and even if specified by the protocol, it is not part of the definition of "token". If we ever choose to support non-opaque tokens, there's no reason to change this definition.

draft-ietf-gnap-core-protocol.md
```diff
@@ -125,37 +125,27 @@ Roles are defined by the actions taken and the expectations leveraged
...
-
-Resource Server (RS, aka "API")
-: Accepts tokens from the client instance issued by the AS and serves delegated resources
-    on behalf of the RO. There could be multiple RSs protected
-    by the AS that the client instance will call.
+: application operated by an end-user that consumes resources from one or several RSs, possibly requiring access privileges from one or several ASs. 
+
+    Note: this specification differentiates between a specific instance (the Client instance, identified by its public key) and the software running the instance (the Client software). For some kinds of Client software, there could be many instances of a single piece of Client software.
+
+    Example: a Client can be a mobile application, a web application, etc.
```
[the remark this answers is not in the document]
fimbault: Sure will update accordingly.

draft-ietf-gnap-core-protocol.md
```diff
@@ -125,37 +125,27 @@ Roles are defined by the actions taken and the expectations leveraged
 on the role by the overall protocol. 
 
 Authorization Server (AS)
-: Manages the requested delegations for the RO. 
-    The AS issues tokens and directly delegated information to an instance of the client.
-    The AS is defined by its grant endpoint, a single URL that accepts a POST
-    request with a JSON payload. The AS could also have other endpoints,
-    including interaction endpoints and user code endpoints, and these are
-    introduced to the RC as needed during the delegation process. 
+: server that grants privileges to a particular end-user and that provides them to a client in the form of an access token.
```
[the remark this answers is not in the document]
fimbault: We'd have to explain what we mean by "delegated privileges" (that's what the reference to the end-user was trying to circumvent, but maybe it wasn't the best way to do it / I understand what you mean here). 
The rest of your proposal seems fine to be.

draft-ietf-gnap-core-protocol.md
fimbault: There was the question whether the definitions should allow the possibility for several servers. If we want to reflect the current state of affairs, your proposal is better.

draft-ietf-gnap-core-protocol.md
fimbault: yes clearer also

draft-ietf-gnap-core-protocol.md
jricher: Opaqueness to the client is a clear line in terms of roles and responsibilities. The token value is a message from the AS to the RS, which the client carries but is not required to understand. If there is other information the client needs, it needs to be sent separately from the token. This is where the "subject information" and assertions and the like come into play. As soon as we say "the token might not be opaque" we need to define what the format and contents of the token are, and set expectations of what we expect clients to understand within the token in order to function. The fact that the client doesn't have to see inside the token is one of the key benefits of OAuth that we would be remiss to throw out here.

Note that the token is only opaque to the client, it is not opaque to the RS or AS.

draft-ietf-gnap-core-protocol.md
fimbault: same remark as before. You're right, but that issue shows we need to explain the relationship between the user and the software (aka delegated access).

draft-ietf-gnap-core-protocol.md
fimbault: No, but then we probably should always refer to protected resource everywhere?

draft-ietf-gnap-core-protocol.md
fimbault: good catch

draft-ietf-gnap-core-protocol.md
fimbault: editorial change accepted

draft-ietf-gnap-core-protocol.md
fimbault: editorial change accepted

draft-ietf-gnap-core-protocol.md
yaronf: This is all good, I just think it is inappropriate in a *definition*. It's an important property of the token, yes, but it's not part of its definition IMHO.

fimbault: Yes, we tried to avoid having definitions that embed requirements (here opaque). We should put that in another section.

draft-ietf-gnap-core-protocol.md
jricher: Yes, it probably should either be protected resource or, most likely, RS everywhere as that's what the protocol deals with.

draft-ietf-gnap-core-protocol.md
fimbault: proposal : a data artifact representing a set of rights and/or attributes.

again we have the "delegated" part which would need a definition if we want to include that. I'll think about it.

draft-ietf-gnap-core-protocol.md
jricher: My proposal :allows: several servers without, in my opinion, confusing the matter.

draft-ietf-gnap-core-protocol.md
jricher: Oh good catch, I missed that this one was previously RO + RQ and over-corrected.

draft-ietf-gnap-core-protocol.md
fimbault: people seem to have different options about that. Right now I left it generic enough, but this could change in the future if we see we don't need to leave that open.

draft-ietf-gnap-core-protocol.md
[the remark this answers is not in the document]
jricher: If we define "privileges" then this is "privileges that have been delegated to the parties named" without need for further definition, I think. It's tricky though.

draft-ietf-gnap-core-protocol.md
fimbault: ok with that.

draft-ietf-gnap-core-protocol.md
fimbault: currently, it's indeed used only in subject information. The idea was to say that it is a statement made by someone, not a fact. Can merge into the main definition if we don't find that useful.

draft-ietf-gnap-core-protocol.md
fimbault: yes same remark as before, to be propagated everywhere.

draft-ietf-gnap-core-protocol.md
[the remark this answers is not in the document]
fimbault: I keep delegated privileges for the commit, will see later if we can find a better proposal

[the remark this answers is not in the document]
fimbault: I propose not to include "The AS determines which policies apply to any given delegation request." because it's already a sort of requirement, not a definition per say.

draft-ietf-gnap-core-protocol.md
fimbault: ok, proposed change: 

Resource Server (RS)
: server that provides operations on protected resources where operations require a valid access token issued by an AS trusted by the RS for this purpose.

+ keep protected resource definition
- remove resource definition

Will have to check in the main text that we refer to the RS or protected resources

draft-ietf-gnap-core-protocol.md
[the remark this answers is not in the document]
jricher: OK, I see where you're going with the thinner definitions now. We can add in the requirements in a different part.

draft-ietf-gnap-core-protocol.md
fimbault: When doing it,this creates long/convoluted sentences -> I just add a note "Note: to avoid complex sentences, the specification document may simply refer to resource instead of protected resource."

draft-ietf-gnap-core-protocol.md
fimbault: Currently left a note in issue #145, resolving this conversation for now

draft-ietf-gnap-core-protocol.md
fimbault: To keep that generic, I suggest 
"right or attribute associated with a subject."

I also included a note
Note: the RO defines and maintains the rights and attributes associated to the protected resource, and might temporarily delegate some those privileges to an end-user. This process is refered to as privilege delegation. 

-> as a proposal to solve the delegated privilege issue

draft-ietf-gnap-core-protocol.md
fimbault: removed

draft-ietf-gnap-core-protocol.md
fimbault: Proposal: (verb): to permit an instance of client software to exercise some set of delegated rights to access a protected resource and/or to receive some attributes at a specific time and during a specific duration. (noun): the act of granting.

draft-ietf-gnap-core-protocol.md
fimbault: I didn't include "trusted by the RS for this purpose." because I don't want a circular definition

fimbault:
A new commit has been made, and integrates the changes based on Justin's feedbacks. 
Wiki updated too.

jricher (approved):

draft-ietf-gnap-core-protocol.md
```diff
@@ -204,29 +194,35 @@ role of the AS as defined by the protocol.
...
-and verified by the RS, and issued to and carried by the client instance. The contents
-and format of the access token are opaque to the client.
+: a data artifact representing a set of rights and/or attributes.
+
+    Note: an access token can be first issued to an client instance (requiring authorization by the RO) and subsequently rotated.
 
 Grant
-: The process by which the client instance requests and is given delegated
-access to the RS by the AS through the authority of the RO.
+: (verb): to receive some attributes at a specific time and valid for a specific duration and/or to permit an instance of client software to exercise some set of delegated rights to access a protected resource (noun): the act of granting.
```
jricher, line 206: This breaks the intended verb, which was "permit" for two actions of the client, to exercise and to receive. This should be instead:

```suggestion
: (verb): to permit an instance of client software to receive some attributes at a specific time and valid for a specific duration and/or to exercise some set of delegated rights to access a protected resource (noun): the act of granting.
```

fimbault: yes sorry, the problem with copy paste. Should be fixed now with latest commit.

PR #160, opened 2021-01-13
Clean up changelog, add missing items

jricher:


fimbault (approved):
This seems fine, indeed it would be better that we change that part systematically at each PR.

Likewise, at some point we'll have to cleanup the list of issues in the document itself, such as [[ See issue #29 ]] which was closed. But maybe this will be more practical to do when we feel we have a document nearly ready for approval (so not now).

Issue #161, opened 2021-01-20
Terminology & lifecycle: revoke / rotate


fimbault:
So I was starting to have a look at rotation/revocation issues (keys or tokens), as we have many open issues that relate to that

Interesting insight from DIF KERI (Universal Identifier Theory paper)
"The term revocation has two completely different meanings in the identity space. In key management one may speak of revoking keys. With statement issuance, authorization issuance, or credential issuance, one may speak of revoking an authorization statement, a token, or a credential. This becomes confusing when the act of revoking keys also implicitly revokes the authorization of statements signed with those keys."

and "KERI terminology usually avoids that confusion between rotation and revocation because a key rotation operation is the equivalent of a key revocation operation followed by a key replacement operation. So one operation, rotate, is implemented instead of two operations (revoke and replace)."

fimbault:
Related question: what is persistent ? what is ephemeral ? 

And more specifically:
- every new request needs to provide a key -> don't need long term persistance for that, we simply need to handle key rotations during a single request (as described in issue #85) 
- objets derived from some key however (including bound tokens) might be more tricky. Do we need some kind of registry to keep track of the underlying changes? Or do we just consider that they get revoked?

Issue #37, opened 2020-11-13
"split_token" functionality and flag


jricher:
§2.1.4 Signaling Token Behavior: Editor's note: 

This functionality is controversial at best as it requires significantly more complexity on the client in order to solve one class of AS/RS deployment choices.

Issue #14, opened 2020-11-01
Split Tokens


yaronf:
**Yaron** (referring to `split_token`): I suggest to extend the editor's note to clarify what is the use case, so that the group can decide whether split tokens are indeed necessary.

**Justin**: I think we should have a more detailed conversation about this potential feature than an editor’s note could reasonably get into. There are a number of large items (splitting access tokens, directing access tokens to specific portions of API or RS’s, among others) that I think will be big conversations for the WG, and all the context of these don’t necessarily need to be documented inside the notes in the document itself. 

But for discussion and context, this feature was proposed to deal with cases where the AS might decide to issue several access tokens in any case where the client requested one, because the AS wants to have restricted audiences for individual access tokens but the client doesn’t know about the audiences ahead of time. I agree that it’s a compelling use case, and it puts security of the protected domains in the hands of the AS where it belongs, but it would require a client to be smart enough to know what to do with these tokens downstream, and if OAuth has taught us anything it’s that we can’t rely on client software to be smart about most anything. :)

jricher:
See also #37

adeinega:
I always considered the "split_token" functionality as a nice addition to the protocol. It gets available for the client when it explicitly requests it + the AS can issue multiple ATs.

Among other described scenarios above, I had in mind this particular one; let's say there are two almost identical RSs, the difference is only in the way how much you need to pay to use them, for example, our client wants to store video recordings, RS2 works a bit faster and guarantees more protection against data loss. At the same time, RS1 and RS2 belong to different organizations that have nothing in common + the client knows anything about RS1 and RS2 in advance.

jricher:
@adeinega For this scenario, you'd also need a way to explicitly direct the client instance to either RS1 or RS2 along with the token. There's nothing in the protocol that does that today (see #69 ). Both of these could be handled as an extension.

The current view of the editors is that this is a "nice to have", which isn't enough for it to be in core, especially not without implementation experience.

adeinega:
[...]

PR #162, opened 2021-01-25
Reformat access token request and response format.

### LINKED ISSUE #10: Single vs. Multiple Access Tokens
**Yaron**: an array for a single AT vs. an object for multiple ATs? This is extremely non-intuitive and probably ugly to code. Also, why should (multiple) tokens be named if a single token doesn't have to, why not list them in an array instead?

**Justin**: I agree that it looks a little odd at first glance, but I’ve coded this in a few languages now and on the request side (this section) it is really clear what you’re doing. I think it might be helpful to build up from the parts of this section of the request, which will show the history and decision process that got us to the syntax that’s there today. Obviously, this isn’t a statement that it needs to stay this way, but I hope that this will help bring to light the reasoning behind what’s there now so that it can be explained, refined, or updated as needed.

At the lowest level, we’ve got the dimensions for access. OAuth 2 showed us that “scope” was a useful way for a client to indicate what it wants, but we’ve found that API designers have used scope to talk about many different dimensions of access, all rolled into different scope values. We are now being explicit about those dimensions, and trying to make those dimensions orthogonal to each other. There is some small amount of fuzziness and potential overlap of interpretation, and ultimately these are all up to the AS/RS (and the API designer, really) to define:

 - type: “the kind of API I want to call”, like OIDC UserInfo or FHIR Medical Record
 - datatypes: “the kind of data I want to use at the API”, like email addresses or office visits
 - actions: “the kind of actions I want to do on the API”, like read, or write, or make the robot move
 - locations: “the RSs I want to call the API on”, like the URL of the UserInfo Endpoint or a URI for a specific document server cluster
 - identifier: “the entity I want to do these actions on”, like the medical record ID or bank account number

These are all strings or arrays of strings, and they’re meant to be very specific to the requested access to the API. Specific APIs are likely to add their own dimensions as well, which can be pretty much anything so long as the RS can interpret the token that it represents. This is why we have the “type” field, which gives us a way to namespace the dimensions. 

But these dimensions don’t live alone: I’m not going to just ask for “read” or just “medical record”, I probably want to ask for “read the medical record”. With scopes, we were forced to just put them all in a bucket and define awkward semantics for combining them, and then watch things go awry as an AS protected different APIs with different interpretations of, for example, “email” as a scope. So with GNAP, we’ve got a way to combine them into single objects. That lets us create combinations like “read the medical record” by combining the “type” and “actions” fields into a single object. The “pass by reference” feature allows us to send a simple string value that is interpreted as if it were one of the objects. So maybe there’s a “fhir-read” value that means “read all available sections of the fhir medical record of the current user”. You can think of it like a pre-computed object where the client doesn’t have to specify all the details — and in that way it works almost just like a scope value would. 

But what if I want read access to the medical history record but ALSO want update access to the demographic information? We can’t just throw it all onto the same object because there are two very distinct requests, but one access token can reasonably encompass both. So in GNAP we’ve got an array of objects that allows us to combine these multidimensional requests and have them all apply to a single access token. Since any of these objects could potentially be replaced by a string, this array could have a combination of strings or objects, but each item in the array represents the same kind of multi-dimensional request. The “scope” field in OAuth 2 is semantically an array (even though it’s formatted as a space-separated list of space-free ascii strings) for exactly this reason: I should be able to ask for multiple scopes at the same time. 

The resulting access token is for the total of everything in the array (that’s granted — of course any of this can be changed by the AS and RO). So an access token always represents an array of requested access rights, each access right is an object that describes what’s being requested across several dimensions which are properties of that object in their own syntax. An access token is a combined collection of access rights, so requesting an access token requires the structure of an array. And so far, this is inline (by design) with what’s in OAuth 2 RAR:  https://tools.ietf.org/id/draft-ietf-oauth-rar-03.html

And that brings us to multiple access tokens: Multiple access tokens are requested by sending an object which uses labels to mark individual single access token requests, which as we just saw is an array — so it’s an object whose values are all arrays, but specifically arrays that are defined to ask for a single access token.

Why put the labels on these requests? Multiple access tokens need to have names (or labels) so that the client can tell :which: of its access requests each token maps to. Otherwise, how can a client know which set of resources it’s supposed to use each token with? The “locations” field (discussed more below) is for identifying sets of resources, and so that field is not likely to be the exact set of URLs that the client can use the token at. And it’s also not guaranteed that all APIs will use “locations” in a way that’s meaningful to a client — what if the client wants “read” and “write” access at the same API with different tokens? The “read” token could be long-lived but the “write” token short lived, for security purposes, but the “locations” would be the same. Additionally, the “read” could be limited to GET requests and the “write” limited to “POST” requests, neither of which is really expressed in “locations”. If the RC is going to ask for different tokens in a single request, it needs to have a way to differentiate the portions of the response from each other. 

There are some different syntaxes we could use to make parts of this work, but these choices tend to have other follow-on effects. Here are a few that I’ve cycled through in the design of the current proposal:

We could ask for multiple access tokens this by having a “label” field inside the request for a single access token and then sending an array of single access token request (so an array of arrays), but then you need to answer the question of where in the request to put the label. Should it go in the array like the proposed “token flags”? Should it go inside one of the objects in the array? But if it does, what would it mean for multiple objects in the array to have a “label” value? Additionally, we’d need to detect at the top level whether we had an array of arrays or an array of objects and strings — doable, but more prone to weird error cases. (What if an RC sends an array with one array and one object? We need to detect that now and might end up down the wrong code path before we do, with a naive parser anyway.)

We could simply require all tokens to be labeled at all times, but I would argue that the overwhelmingly common use case of a single access token doesn’t need that kind of overhead because there’s only one set of requested items that the RC is asking for. 

We could define single access tokens as an object, with multiple aspects. This would potentially let us hang additional fields off of the access token request alongside the resource list, but it would also mean that we’d have a degenerate case where you’re defining an object with one member, whose value is always an array. In other words, this is exactly what the JSON Web Key Set format does, which I’ve always found clunky. The top level object doesn’t add anything in the JWKS case, and I’m not seeing how it would add anything here. On the other hand, this would let us use an object for a single access token and potentially an array for multiple access tokens, but at some cost to the complexity at the layers below that.

None of those ideas lead to a better syntax across the different uses cases, or that optimize for simplicity in the common use cases.

All that is to say that I currently think that what we have is the simplest and cleanest approach that I’ve seen so far, and I would really love to see if there are any alternatives that offer this same kind of power and clarity when exercised the same. I would love to see some fresh ideas on different ways to approach this.

**Yaron**: wow, that was a long read, thank you! IMHO an array (which in the common case has a single element) is just fine. I hate making up names ("labels") for no good reason.

### LINKED ISSUE #13: Syntax of Flags
**Yaron**: why are flags represented as resources? The answer, "this is how OAuth 2 does it" is not great.

**Justin**: The idea is that these are items that alter what the access token that’s being requested in the “resources” section. It is a sensible way to hang things together since everything in the array ties to a specific access token. This was a proposed design compromise, and if these flags can fit elsewhere then we can do that.

**Yaron**: this still doesn't explain why we don't have a `flags` object (or array), parallel to the `resources` array.

### LINKED ISSUE #39: Token flags
§2.1.4 Signaling Token Behavior: Editor's note: 

While these reference values are "reserved", the ultimate decider for what a reference means is the AS, which means an AS could arguably decide that one of these values means something else. Also, this kind of reservation potentially steps on API namespaces, which OAuth 2 is careful not to do but common extensions like OIDC do with their own scope definitions. However, in OIDC, several "scope" values have behavior similar to what's defined here, particularly "openid" turns on ID tokens in the response and "offline_access" signals for the return of a refresh token, and these can be used outside of OpenID Connect itself. However, to keep these flags out of the general API namespace, we could use a different syntax for sending them. In particular, they could be defined under a GNAP-specific "type" object, where all the flags are fields on the object.

```
resources: [
    {
        type: "gnap-flags",
        flag1: true,
        flag2: false,
        flag3: true ...
    },
    "reference1",
    "scope2", ...
]
```

Alternatively, all the flags could be sent in an array separate from the rest of the request.

```
resources: [
    "reference1",
    "scope2",
    ["flag1", "flag2", "flag3"] ...
]
```

This whole thing might also belong in an extension, as it's advanced behavior signaling for very specific cases. However, it seems other extensions would be likely to extend this kind of thing, like OIDC did with "offline_access".

### LINKED ISSUE #40: Access token request format
Currently, access tokens are requested using the `resources` parameter, but there might be different things to affect the resulting access tokens, such as the token flags in 2.1.4. 

An alternative would be to allow tokens to be requested using an object syntax where `resources` is one of the elements:

```
"access_token": {
    "key": true,
    "split": true,
    "label": "token-1",
    "resources": [
        {
            "type": "photo-api",
            "actions": [
                "read",
                "write",
                "dolphin"
            ],
            "locations": [
                "https://server.example.net/",
                "https://resource.local/other"
            ],
            "datatypes": [
                "metadata",
                "images"
            ]
        },
        "read"
    ]
}
```

In this example, the `key` field mimics what the `key` field of the token response does, here indicating that the token should be bound to the client instance's own key. The `split` field takes the place of the `split_token` flag (and ultimately we wouldn't need the token flags structure). The `label` field is a client-supplied token label, which would be required if the client requested multiple tokens, such as something like this:

```
"access_token": [
  {
    "key": false,
    "label": "reader",
    "resources": [ "read" ]
  },
  {
    "key": true,
    "label": "writer",
    "resources": [ "write ]
  },
]
```

This would take the place of the multiple-token syntax map used in §2.1.3 today.

jricher:
Closes #40
Closes #39
Closes #10
Closes #13

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4799,9 +4799,7 @@ Content-type: application/json
 ~~~
 
 
-
-
-# No User Involvement {#example-no-user}
+## No User Involvement {#example-no-user}
 
 In this scenario, the client instance is requesting access on its own
 behalf, with no user to interact with.
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -564,7 +564,7 @@ The client instance polls the AS while it is waiting for the RO to authorize the
 
 3. The AS determines which RO to contact based on the request in (1), through a
     combination of the [user request](#request-user), the 
-    [resources request](#request-resource), and other policy information. The AS
+    [resources request](#request-token), and other policy information. The AS
     contacts the RO and authenticates them.
 
 4. The RO authorizes the pending request from the client instance.
@@ -678,10 +678,9 @@ To start a request, the client instance sends [JSON](#RFC8259) document with an 
 member of the request object represents a different aspect of the
 client instance's request. Each field is described in detail in a section below.
 
-resources (object / array of objects/strings)
-: Describes the rights that the client instance is requesting for one or more access tokens to be
-    used at RS's. {{request-resource}}
-   
+access_token (object / array of objects)
+: Describes the rights and properties associated with the requested access token. {{request-token}}
+
 subject (object)
 : Describes the information about the RO that the client instance is requesting to be returned
     directly in the response from the AS. {{request-subject}}
@@ -714,25 +713,27 @@ A non-normative example of a grant request is below:
 
 ~~~
 {
-    "resources": [
-        {
-            "type": "photo-api",
-            "actions": [
-                "read",
-                "write",
-                "dolphin"
-            ],
-            "locations": [
-                "https://server.example.net/",
-                "https://resource.local/other"
-            ],
-            "datatypes": [
-                "metadata",
-                "images"
-            ]
-        },
-        "dolphin-metadata"
-    ],
+    "access_token": {
+        "access": [
+            {
+                "type": "photo-api",
+                "actions": [
+                    "read",
+                    "write",
+                    "dolphin"
+                ],
+                "locations": [
+                    "https://server.example.net/",
+                    "https://resource.local/other"
+                ],
+                "datatypes": [
+                    "metadata",
+                    "images"
+                ]
+            },
+            "dolphin-metadata"
+        ]
+    },
     "client": {
       "display": {
         "name": "My Client Display Name",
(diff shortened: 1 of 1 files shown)


aaronpk (approved):
Looks good to me!

fimbault (approved):
Seems fine to me, thanks. I copy the answers to my questions, as I think they're useful to understand some of the choices.

- does it mean we make token binding the default behaviour? (if so we should explicit why this choice is made)
Yes, that is the idea. One of the goals of GNAP is to have it more secure than OAuth2 by default, and having bound tokens be first class and relatively simple (ie, it’s just token presentation and key proof combined, and you need to do key proof anyway) is in line with that goal. 
I don't think there's a need to be more explicit in the spec itself but at least it clarifies the intent.

- I'm also wondering whether the split flag is to be provided on the client side. Do we think there could be reasons for clients not to be able to deal with that?I would have expected that this is more something driven by the AS, based on policies the client might want to be able to be aware of (through OPTION). 
- Splitting tokens is an advanced piece of functionality that drastically increases the complexity on the client side, and so I would only want this to be an option if the client chooses it. My design philosophy around things like this in the protocol is that a simple client should be able to implement a subset of the protocol’s syntax and still function just fine, in a predictable and secure fashion.

draft-ietf-gnap-core-protocol.md
```diff
@@ -771,72 +772,66 @@ The request MUST be sent as a JSON object in the body of the HTTP
...
-      define which other fields are allowed in the request object. 
-      This field is REQUIRED.
+split
+: If this flag is included, the client instance is capable of 
+    receiving a different number of tokens than specified in the
+    [token request](#request-token), including 
+    receiving [multiple access tokens](#response-token-multiple)
+    in response to any [single token request](#request-token-single) or a
+    different number of access tokens than requested in a [multiple access token request](#request-token-multiple).
+    The `label` fields of the returned additional tokens are chosen by the AS.
```
yaronf, line 819: We should say explicitly that the `label` is opaque, and must not be used by the AS to denote the usage of access tokens (in the "split" case), otherwise they will do just that.

draft-ietf-gnap-core-protocol.md
```diff
@@ -852,213 +847,76 @@ using the fictitious `photo-api` type definition.
...
-fields specified here as well as the API-specific `geolocation`
-field. The second access request includes the `actions` and
-`identifier` fields specified here as well as the API-specific
-`currency` field.
-
-~~~
-    "resources": [
-        {
-            "type": "photo-api",
+            "type": "walrus-access",
```
yaronf, line 852: What use is this `type`?

draft-ietf-gnap-core-protocol.md
```diff
@@ -1899,54 +1688,85 @@ properties.
...
+
+key (object / string)
+: OPTIONAL. The key that the token is bound to, if different from the
+    client instance's presented key. The key MUST be an object or string in a format
+    described in {{key-format}}, describing a public key to which the
+    client instance can use the associated private key. The client instance MUST be able to
+    dereference or process the key information in order to be able
+    to sign the request.
+
+durable (boolean)
```
yaronf, line 1743: I don't see how this is useful, IOW how the client's code would differ depending on this setting. In either case the access token might suddenly expire, and the client needs to deal with that.

draft-ietf-gnap-core-protocol.md
```diff
@@ -1899,54 +1688,85 @@ properties.
...
+    If this flag is set to the value `true`, then the client instance can expect
+    a previously-issued access token to continue to work after it has been [rotated](#rotate-access-token)
+    or the underlying grant request has been [modified](#continue-modify), resulting
+    in the issuance of new access tokens. If this flag is set to the boolean value `false`
+    or is omitted, the client instance can anticipate a given access token
+    will stop working after token rotation or grant request modification.
+    Note that a token flagged as `durable` can still expire or be revoked through
+    any normal means.
+
+split (boolean)
```
yaronf, line 1754: This is not very useful. If I request "L1", "L2" and "L3", and I get back L1, L4, L5, L6, L7 where everything but L1 is marked as "split", what do I make of the response? Do I try to correlate the 4 access tokens in the response with L2 and L3 in the request? How?

draft-ietf-gnap-core-protocol.md
```diff
@@ -1968,57 +1788,60 @@ URL that has access to three described resources.
...
             "finance", "medical"
         ]
     }
 ~~~
 
 
-If the client instance [requested multiple access tokens](#request-resource-multiple), the AS MUST NOT respond with a
-single access token structure unless the client instance sends the `split_token` flag as described in {{request-resource-flag}}.
+If the client instance [requested multiple access tokens](#request-token-multiple), the AS MUST NOT respond with a
+single access token structure unless the client instance sends the `split` flag as described in {{request-token-single}}.
```
yaronf, line 1806: On the contrary, if the request has the `split` flag, you can expect the response to contain *more* ATs than the request.

You may have wanted to say (or IMO you should say) that if the client requests multiple ATs, the AS must not return a single AT structure, even if only a single AT is returned.

draft-ietf-gnap-core-protocol.md
```diff
@@ -1968,57 +1788,60 @@ URL that has access to three described resources.
...
 If the client instance has requested multiple access tokens and the AS has
 granted at least one of them, the AS responds with the
-"multiple_access_tokens" field. The value of this field is a JSON
-object, and the property names correspond to the token identifiers
-chosen by the client instance in the [multiple access token request](#request-resource-multiple).
-The values of the properties of this object are access
+"access_token" field. The value of this field is a JSON
+array, the members of which are distinct access
 tokens as described in {{response-token-single}}.
+Each object MUST have a unique `label` field, correspond to the token labels
```
yaronf, line 1818: correspond -> corresponding

yaronf (changes requested):

draft-ietf-gnap-core-protocol.md
jricher: The "split token" function is being tracked in other existing issues: #14 and #37, this PR just packages the existing function. Conversation on changing that function can happen on those issues and be addressed with a new PR.

draft-ietf-gnap-core-protocol.md
jricher: Token rotation is being tracked on #161 #101 #96 and a number of other issues. This function was previously an input flag from the client instance, but the editors are here framing it as an output state flag instead because there wasn't a lot of support for having the client signal this going in.

draft-ietf-gnap-core-protocol.md
jricher: Thanks, I think this was a copy-paste error that needs to be cleaned up!

draft-ietf-gnap-core-protocol.md
jricher: The "type" field defines the API used. This structure is defined in section 8 now, but it's the same as it was before: https://deploy-preview-162--gnap-core-protocol-editors-draft.netlify.app/draft-ietf-gnap-core-protocol.html#name-resource-access-rights

yaronf (approved):

PR #165, opened 2021-01-27
prevent PR label check from running on closed PRs

aaronpk:


jricher (approved):
Looks good to me, the `state` check was the magic it was missing, I think. We'll see when this is merged and cleaned up, I guess. :)

aaronpk:
Yep, and I had to ask the author of that plugin we're using, it was going to be a hopeless effort trying to find that in the docs. I challenge you to find a reference to this option even knowing what it's called now. 😉

aaronpk:
I tried removing the label [here](https://github.com/ietf-wg-gnap/gnap-core-protocol/runs/1780036620?check_suite_focus=true) but it failed again, I think because it's running the code from the repo at that point, not from the current head. The real test will be whether PRs created after today work!

Issue #106, opened 2020-11-13
JWS Headers for JOSE-based signature methods


jricher:
§8.1 Detached JWS: Editor's note: 

It's not the usual practice to put additional information into the header of a JWS, but this keeps us from having to normalize the body serialization. Alternatively, we could add all these fields to the body of the request, but then it gets awkward for non-body requests like GET/DELETE. 

§8.2 Attached JWS: Editor's note: 

It's not the usual practice to put additional information into the header of a JWS, but this keeps us from having to modify the body to use this signature method. Alternatively, we could add all these fields to the body of the request, but then it gets awkward for non-body requests like GET/DELETE.

fimbault:
This is a design choice. What can't be done or what is made difficult by what is currently in the text? Please advise or will be closed.

davidgtonge:
Needs to be referenced in #15 - there are definite trade-offs with the different signature options and they need to be considered

Issue #98, opened 2020-11-13
Reading the state of a request


jricher:
§5.4 Getting the Current State of a Grant Request: Editor's note: 

I'm a little dubious about the need for this particular function in reality, but including it for completeness sake. There are a lot of questions we need to answer, such as whether it's safe to include access tokens and claims in the response of this kind of "read" at all, and whether it makes sense to include items like interaction nonces in the response. This discussion should be driven by the use cases calling for this "read" functionality. There have been similar functions within proprietary protocols where the client calls an endpoint at the AS to figure out where the user is in the interaction process at the AS, letting the client provide a smarter UI. It doesn't seem like we could do that in depth here since it would be highly application specific, but that might be a good example of how to extend a response and give a client extra information.

jricher:
Editors suggest this section requires additional text to explain why this action is needed and under what circumstances. Without additional text, this section will be removed in a future update.

fimbault:
Closing the issue following the merge of the PR

Issue #55, opened 2020-11-13
Unique callback URIs


jricher:
§2.5.3 Receive a Callback After Interaction: Editor's note: 

Should we enforce the callback URI to be unique per request? That helps with some fixation attacks, but not with others, and it would be problematic for an AS that wants to lock down each client instance to a single callback instead of a family/pattern of callbacks.

jricher:
Editors suggest closing as there are valid reasons for an AS to require an exact match of a single callback URI, such as establishing a client's identity without any pre-registration.

dickhardt:
What are the attacks that are still possible with a unique callback URL?

AS already lock down from a family or URLs. Or the callback could be a fixed URL with a parameter that makes the URL unique.

aaronpk:
The question in this issue is whether to *require* unique URLs, or whether the current text that allows stable URLs as well as unique URLs is okay.

Stable redirect URLs have a number of advantages, such as being able to be compared with exact string matching rather than needing to parse the URL and compare host/path/query/etc components. If the AS needs to be able to parse the URL, that opens up a whole world of [possible SSRF attacks](https://www.blackhat.com/docs/us-17/thursday/us-17-Tsai-A-New-Era-Of-SSRF-Exploiting-URL-Parser-In-Trending-Programming-Languages.pdf) that are avoided with simple string matching.

PR #170, opened 2021-02-08
closing issue 98, effectively removing section 5.4

fimbault:
- Reading the state of a request (#98) - removes section 5.4
- Updating and reading access tokens (#99) 
- Remove the links to already closed issues in the spec (<strike>issues #29, #55, #106</strike>, #98, #99)
- Update the list of high level changes for upcoming version 4 in appendix A

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -182,6 +182,8 @@ another system that deals with security rights. From the perspective of
 GNAP, all of these are pieces of the AS and together fulfill the
 role of the AS as defined by the protocol.
 
+\[\[ [See issue #29](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/29) \]\]
+
 ## Elements {#elements}
 
 In addition to the roles above, the protocol also involves several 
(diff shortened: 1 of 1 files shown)


yaronf:
Guys, please don't mix editorial and technical commits in the same PR. 😢

jricher:
@yaronf this does not mix those types. This is a removal of a section that was marked for removal several months ago and has received no input to keep it, after several requests by the editors.

jricher:
On second read, I realize I missed a change and I agree with Yaron: this PR should not remove issue links to unrelated issues. Those should be in a separate PR.

jricher (changes requested):
This PR should only close the links for only the affected issues. A cleanup PR should be added separately as needed.

fimbault:
Change done, only 98 and 99 are now concerned.

draft-ietf-gnap-core-protocol.md
```diff
@@ -4301,6 +4273,7 @@ sure that it has the permission to do so.
 - -04
     - Updated terminology.
     - Refactored key presentation and binding.
+    - Closed issues related to reading and updating access tokens
```
jricher, line 4276:
```suggestion
    - Removed function to read state of grant request by client
    - Closed issues related to reading and updating access tokens
```

jricher (changes requested):
just needs the changelog updated (included) and otherwise looks good

draft-ietf-gnap-core-protocol.md
fimbault: yes that's clearer, done

jricher (approved):

davidgtonge:
I'm a bit late as this has already been approved. I'm ok with the text being removed, but I do think we need to address https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/146

PR #177, opened 2021-02-13
Fix subject identifier type hint

adeinega:
A Subject Identifier is a JSON object, not a string per https://www.ietf.org/id/draft-ietf-secevent-subject-identifiers-06.txt.

https://www.ietf.org/id/draft-ietf-secevent-subject-identifiers-06.txt#:~:text=A%20Subject%20Identifier%20is%20a%20JSON,identify%20a%20principal%20within%20some%20context.


jricher (approved):
Good catch, thanks!

PR #179, opened 2021-02-14
Fix typos.

adeinega:


draft-ietf-gnap-core-protocol.md
```diff
@@ -1626,7 +1626,7 @@ uri (string)
             value exactly as given when making a [continuation request](#continue-request).
 
 wait (integer)
-: RECOMMENDED. The amount of time in integer
+: RECOMMENDED. The amount of time in
```
jricher, line 1629:
```suggestion
: RECOMMENDED. The amount of time in integer
```

I think this can remain "integer" here to indicate this doesn't allow for fractional seconds. While the type hint above is set to "integer", it doesn't hurt to say it again inside.

jricher (approved):
One small change needed (marked) but otherwise good!

Issue #180, opened 2021-02-14
grant_request_endpoint, HTTPS and fragment component


adeinega:
Sections "9 Discovery" states that

> grant_request_endpoint (string) REQUIRED. The full URL of the AS's grant request endpoint.

It makes sense to specify that this URL should be using HTTPS. I also don't think you want to allow using the fragment component in it, as well as query parameters. The query parameters could be arguable though...

jricher:
This definition should be beyond just the discovery portion -- this endpoint is fundamental to the protocol and it does need to be protected. We can at least mention that it starts with HTTPS here. Prohibiting fragments makes sense, but prohibiting query parameters or any other URL portion does not make sense, though, as that could be important to an AS's deployment. However, the client instance always takes this string as-is and doesn't add or modify anything, and it always makes a POST to it.

adeinega:
https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/183

adeinega:
I've reused the definition of [the Issuer location from OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html#:~:text=a%20host%20component%2C%20and%20optionally%2C%20port%20and%20path%20components%20and%20no%20query%20or%20fragment%20components). That is a very precise one IMO.

jricher:
closed by #183

PR #183, opened 2021-02-15
Requirements for the location of grant_request_endpoint.

adeinega:
Suggested changes for https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/180

They could be also simplified to something like

> This URL MUST use the HTTPS scheme and MAY contain port, path, and query parameter components. The URL MUST NOT contain fragment components.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -45,6 +45,7 @@ normative:
            ins: P. Saint-Andre
     RFC2119:
     RFC3230:
+    RFC3986:
     RFC5646:
     RFC7468:
     RFC7515:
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -4073,9 +4074,11 @@ containing the following information:
 
 
 grant_request_endpoint (string)
-: REQUIRED. The full URL of the
-          AS's grant request endpoint. This MUST match the URL the client instance used to
-          make the discovery request.
+: REQUIRED. The location of the
+          AS's grant request endpoint. The location MUST be a URI {{RFC3986}}
```
jricher, line 4078: Since this URI has a fixed scheme, it is more properly a URL. I'm not sure if the reference should remain this RFC or point to 3305 or something else?

```suggestion
          AS's grant request endpoint. The location MUST be a URL {{RFC3986}}
```

adeinega: I don't have any preference as to 3305 vs 3986.

Issue #178, opened 2021-02-14
sub_ids and different contexts


adeinega:
It may make sense to use 

1. subject_types (or subject_type_identifiers) instead of sub_ids in section 2.2. Requesting Subject Information

> sub_ids (array of strings) An array of subject identifier subject types requested for the RO...

2. subject_types (or subject_type_identifiers) instead of sub_ids in section 9. Discovery

> sub_ids (array of strings) OPTIONAL. A list of the AS's supported identifiers.

The reason for it is that the same sub_ids has a different meaning in section 3.4. Returning User Information, where it's

> sub_ids (array of objects) An array of subject identifiers for the RO

and so on.

Furthermore, `A list of the AS's supported identifiers.` in section 9. Discovery could be changed to `A list of the AS's supported identifier types.` for clarity.

See also https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/177

jricher:
I like this idea, as these really are two different entities with different purposes.

fimbault:
Makes sense on the general idea. 
Made some comments on the PR regarding the label.

fimbault:
This also questions what could be the types of sub_ids
-> we currently refer to https://www.ietf.org/archive/id/draft-ietf-secevent-subject-identifiers-06.txt
-> but that could also be DIDs (cf #171)
-> or a unique identifier local to the AS: as_uid (suggestion to be used as an example instead of email, cf #16), this would be opaque to other systems but enough for a client to distinguish users (just like using a database identifier)

PR #181, opened 2021-02-14
Fix the example of Attached JWS

adeinega:
Field "htm" should be in uppercase and field "htu" provide the whole HTTP URI.

[...]

PR #184, opened 2021-02-16
sub_ids and different contexts

adeinega:
Suggested changes for https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/178

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -33,7 +33,7 @@ author:
 
 normative:
     BCP195:
-       target: 'https://www.rfc-editor.org/info/bcp195'
+       target: 'http://www.rfc-editor.org/info/bcp195'
        title: Recommendations for Secure Use of Transport Layer Security (TLS) and Datagram Transport Layer Security (DTLS)
        date: May 2015
        author:
@@ -45,7 +45,6 @@ normative:
            ins: P. Saint-Andre
     RFC2119:
     RFC3230:
-    RFC3986:
     RFC5646:
     RFC7468:
     RFC7515:
@@ -205,12 +204,12 @@ Privilege
     Note: the RO defines and maintains the rights and attributes associated to the protected resource, and might temporarily delegate some set of those privileges to an end-user. This process is refered to as privilege delegation. 
 
 Protected Resource
-: protected API (Application Programming Interface) served by an RS and that can be accessed by a client, if and only if a valid access token is provided.
+: protected API (Application Programming Interface) served by a RS and that can be accessed by a client, if and only if a valid access token is provided.
 
     Note: to avoid complex sentences, the specification document may simply refer to resource instead of protected resource.   
 
 Right
-: ability given to a subject to perform a given operation on a resource under the control of an RS.
+: ability given to a subject to perform a given operation on a resource under the control of a RS.
 
 Subject
 : person, organization or device.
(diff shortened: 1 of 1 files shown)


fimbault:
Not sure "subject_type_identifiers" is best (when I read that, what my brain remembers is identifiers, and that's the opposite of what #177 is targeting). 

In 2.2 this is already encapsulated within a "subject" object. In section 9, that's not the case (so we need to explicit that it relates to sub / subject) 
-> "sub_ids_types" or "sub_types"?

adeinega:
@fimbault, I tried to be consistent with https://www.ietf.org/archive/id/draft-ietf-secevent-subject-identifiers-06.txt and called it _subject_type_identifiers_, not "sub_typ_ids" or somehow else... even though, as you know, there is a clear tendency to use short names, e.g. "sub", "kid" and so on and so forth.

IMO, OpenID Connect (in general) and OpenID Connect Discovery (in particular), very consistent with naming and uses "subject_types_supported" and we may also consider using it. What I mean is
1. subject_types_requested for section 2
2. subject_types_supported for section 9

fimbault:
Yes I like that

adeinega:
I've replaced _subject_type_identifiers_ by _subject_types_supported_ which also means is that if you wish to go this route, you would also need to replace at least

1. _key_proofs_ by _key_proofs_supported_
2.  and _assertions_ by _assertion_supporrted_

in section Discovery for the sake of consistency.

These changes, as well as ones in https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/181, aren't ideal and, I believe there will be lots of discussions about subject type identifiers. This is just one little step towards clarity.

fimbault:
After reviewing with the editors, we suggest that we use "subject_types" for section 2 (to be close to secevents) and "subject_types_supported" for section 9.

If additional changes for names are suggested for consistency (keys, assertions), a separate PR is probably best for tracking changes

adeinega:
I've added these changes and created a separate https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/194 Discovery.

draft-ietf-gnap-core-protocol.md
```diff
@@ -4094,9 +4094,9 @@ key_proofs (array strings)
           values of the `proof` field of the 
           [key section](#key-format) of the request.
 
-sub_ids (array of strings)
+subject_types_supported (array of strings)
 : OPTIONAL. A list of the AS's supported
-          identifiers. The values of this list correspond to possible values
+          identifier types. The values of this list correspond to possible values
```
jricher, line 4099: To be particularly pedantic:

```suggestion
          subject identifier types. The values of this list correspond to possible values
```

jricher:
I just realized that this will also need an updated entry in the Changelog since it's a substantive change.

adeinega:
https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/184/commits/a6503423d5c600d3ca612d6760e3551f28c37a6c also fixes a couple of other examples I haven't noticed initially.

jricher:
@adeinega This PR is fine, but in the future, please do not add additional edits after the editors have marked it "Pending Merge". If it's not ready we can back it out of that state, we just want to make sure any WG members have a chance to see the changes as they go in. Thanks!

fimbault:
since SECEVENT has now called the types "format", just wondering if that should be aligned

jricher:
@fimbault I saw that also, and I think we should do a sweep to align them all in one go. My read of the response to this PR is that there's support for changing the terms in this direction (adding _supported and the like). So we might want to hold off on this PR specifically and do all the changes at once. This would affect things I have in #211 and others, so the editors basically just need to rack and stack these to get things fixed throughout the document.

jricher:
(alternatively we can fold in this PR just to change things again with another PR that aligns the terms, which is also reasonable)

adeinega:
Yes... it's quite interesting that the most recent https://datatracker.ietf.org/doc/html/draft-ietf-secevent-subject-identifiers-07 started to use `format` instead of `subject_type`. I believe GNAP should be aligned with _Subject Identifiers for Security Event Tokens_ but it also means it needs to use something a bit more descriptive than something like Discovery parameters' `format_types_supported` or more lengthy `subject_identifier_formats_supported`.

Who knows what next versions of draft-ietf-secevent-subject-identifiers would use in the future? `fmt`?

agropper:
I'm trying to understand how we're using sub_id. See https://mailarchive.ietf.org/arch/msg/txauth/0o3lPPADL5AY84XmqEsWsxU4j2E/

fimbault:
@adeinega good remarks. I don't know why they settled for format, but whatever term they choose, I guess we could ask them to stabilize that once and for all, and avoid forcing us do an additional mapping. 
For discovery, I would suggest `subject_formats_supported`?

jricher:
Accepting as is, will update all SECEVENT references together in a future PR.

PR #194, opened 2021-02-17
Suggested changes for Discovery field names

adeinega:
Suggested changes for https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/184#issuecomment-780684100


jricher (approved):
Looks reasonable to me, and is consistent with changes in #184

jricher:
This will also need an updated entry in the Changelog since it's a substantive change.

adeinega:
@jricher, I've updated the Changelog section and resolved all conflicts.

PR #199, opened 2021-02-18
Adding information about the Cache-Control header and its value.

adeinega:
The specification doesn't mention the Cache-Control header, as well as the Content-Type header for the responses. This is the 1st step to add this information.

All examples of responses may also include this header. Right now, it's true for a few examples.

I am unsure if it's necessary to include a bit obsolete header Pragma.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma#content:~:text=Use%20Pragma%20only%20for%20backwards%20compatibility%20with%20HTTP%2F1.0%20clients.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4146,7 +4146,9 @@ token's presentation, such as its intended proofing mechanism and key
 material.
 
 ~~~
+HTTP/1.1 200 OK
 Content-Type: application/json
+Cache-Control: no-store
 
 {
     "active": true,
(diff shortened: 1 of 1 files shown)


jricher:
This seems like it's still in progress, so I've marked it as a draft for now. Would like to see more input on the right behavior around cache control headers here.

adeinega:
There is no place for the Pragma header. This header is actually for a request, not a response, and it’s for backward compatibility in ancient HTTP 1.0. [https://tools.ietf.org/html/rfc7234#section-5.4](https://tools.ietf.org/html/rfc7234#section-5.4) explains what this header does.

There was even [a suggestion to report an errata for RFC 6750](https://mailarchive.ietf.org/arch/msg/oauth/9DdkE2P0RrUZMeZAbdf3NrMfy0w/) about the improper use of this header. 

The draft for OAuth 2.1 (which is currently under development) still suggests using it. Although, I don’t really think it does it for some particular purpose other than historical reasons.

I believe "Cache-Control: no-store” will be sufficient.

adeinega:
@jricher, do you still want me to provide more inputs on this one?

PR #202, opened 2021-02-23
Syntax and schema based normalization for URI in the htu claim.

adeinega:
This requirement almost one by one copied from https://tools.ietf.org/html/draft-ietf-oauth-dpop-02 as it's a quite important nuance in practical usage.

I suggest also renaming `ts` to `iat` (issued at).

Finally, I think that both attached and detached JWS proving possession mechanisms must include the "typ" header parameter with appropriate values.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -45,6 +45,7 @@ normative:
            ins: P. Saint-Andre
     RFC2119:
     RFC3230:
+    RFC3986:
     RFC5646:
     RFC7468:
     RFC7515:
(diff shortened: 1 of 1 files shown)


jricher (approved):

yaronf:
> This requirement almost one by one copied from https://tools.ietf.org/html/draft-ietf-oauth-dpop-02 as it's a quite important nuance in practical usage.
> 
> I suggest also renaming `ts` to `iat` (issued at).
> 
> Finally, I think that both attached and detached JWS proving possession mechanisms must include the "typ" header parameter with appropriate values.

@adeinega Only `htu` normalization is handed by this PR. Could you open issues for your two other comments?

fimbault:
We don't have the problem in the suggested text, but we might want to tell the DPoP authors that theirs links are incorrect (don't point to RFC3986).

adeinega:
Seem like they the right links and these links get broken during the compilation phase by the mmark processor.

Note, they define `htu` as

> The HTTP URI used for the request, without query and fragment parts.

and this specification as

> The HTTP URI used for this request, including all path and query components.

This is a huge difference.

jricher:
@adeinega Oh -- I thought this got changed in one of the other DPoP fixes. It's not an intentional difference. I'll add an issue, there's other DPoP-related stuff we need to clean up.

jricher:
@adeinega Actually I take it back -- I missed the context here. The JOSE-based methods aren't meant to be copies of the DPoP claims, and there's value in including the full URL here. The claim name shouldn't be the same as the DPoP claim, though, to avoid confusion.

PR #204, opened 2021-03-01
Minor typos

yaronf:
Minor typos

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3068,7 +3068,7 @@ key proof together.
     request is used only for calls to the RS, and only with access tokens that are
     not bound to any key as described in {{response-token-single}}.
 - When neither an access token nor key proof are used, this is an unsecured request. This 
-    type of used only for calls to the RS during a discovery phase as
+    type of request is used only for calls to the RS during a discovery phase as
     described in {{rs-request-without-token}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3881,7 +3881,7 @@ using the fictitious `photo-api` type definition.
 The access requested for a given object when using these fields 
 is the cross-product of all fields of the object. That is to 
 say, the object represents a request for all `action` values listed within the object
-to be used at all `locations` values listed within the object for all `datatype`
+to be used at all `location` values listed within the object for all `datatype`
```
jricher, line 3884: The parameter name is plural, but we can reword the sentence so it reads better:

```suggestion
to be used at all values listed in the `locations` parameter for all `datatype`
```

yaronf: I was trying to be consistent with how the other parameters are named in this paragraph, i.e. plural->singular.

jricher: Oh good catch -- in that case, it looks like the whole paragraph needs to be rewritten to make the parameter names in the descriptions match up with the actual parameter names. I hadn't noticed the inconsistency before.

jricher: We'll take this change now and clean this section up with #205.

Issue #79, opened 2020-11-13
Error responses


jricher:
§3.5 Error Response: Editor's note: 

I think we will need a more robust error mechanism, and we need to be more clear about what error states are allowed in what circumstances. Additionally, is the "error" parameter exclusive with others in the return?

adeinega:
As a starting point, I would suggest adding these sentences

The error code parameter MUST be required and its value SHOULD be limited to ASCII characters.

The error code `invalid_request` indicates that a request is malformed.

The AS MUST return an HTTP 401 (Unauthorized) status code to indicate that it can't identify a client instance (the error code can be `invalid_client`).

The AS MUST return an HTTP 401 (Unauthorized) status code to indicate that it can't authenticate a client instance (the error code can be `invalid_client`).

The AS MAY uncover additional information about the cause of the error when it can authenticate a client instance (the error code can be `invalid_grant`, `invalid_interaction`, and so on).

jricher:
Addressed by #385, still need error codes for things like token management.

adrianhopebailie:
The type of the of the error property in a response is currently ambiguous.
In section 3 it is defined as an `object` and in section 3.6 it is defined as a `string`.

jricher:
@adrianhopebailie thanks for catching that -- looks like both "error" and "error_description" should be added to the list of parameters. I think the intent might have been for the value of "error" to be an object with "error_code" and "error_description" fields inside of it though, so it might be better to align things in that direction instead. Do you have thoughts on which is preferable? 

Interop testing of error conditions is really difficult in practice, so this is a great catch!

PR #206, opened 2021-03-05
HTTP status codes and responses

adeinega:


jricher:
In general I think we're holding off on defining error codes until the core protocol is more solid, but thanks for this! See #79 for the issue tracking this major hole in the spec.

adeinega:
I was a bit surprised that the draft says nothing about any HTTP status codes in case of errors. 400 (and 401) are the most important ones in my opinion. These changes just a little step in this direction... without defining detailed error codes.

jricher:
Not accepted, will address both error codes and response codes in a more comprehensive edit in the future.

Denisthemalice:
I concur with **adeinega** when he says:

> I was a bit surprised that the draft says nothing about any HTTP status codes in case of errors.

Common error codes should be mentioned in the draft, in particular 401, 403 and 405.

This will augment the size of the document of about one page, but it is worth to do it, in particular 
to make the difference between an **_authentication error_** (401) and an **_authorization error_** (403).

jricher:
The editors have agreed (and stated above) that error codes need to be added to the draft. However, that should be done in a more comprehensive manner than what is described in this PR and so we will address it in a future update. Nobody is saying that we should not add error codes, so please don't imply otherwise.

adeinega:
GNAP is closely tied with The Hypertext Transfer Protocol (HTTP) and, as a starting point, it worth specify that

1. 200 (OK) indicates that the request succeeded
2. 400 (BAD REQUEST) indicates that the request hasn't succeeded

before we go any further and identify detailed error codes and this is exactly what I suggested doing in this PR. Although, those are pretty minor changes and if the editors / WG decide to not include them, I'm OK.

This specification could mention lots of other nuances in this area... such as redirect policies (if there are any) for the grant, introspect, continuation, as well as other endpoints. The use and presence of standard response headers like `Date`, `Content-Length` and `Content-Encoding`, an HTTP 429 (Too Many Requests) status code as a way to control the rate of requests, the HTTP Retry-After header, and so forth. The list goes on...

Issue #201, opened 2021-02-22
Include body hash in Detached JWS


aaronpk:
As I was working on a client implementation, it occurred to me that it would be a lot simpler and less error prone to include a hash of the request body in the JWS rather than doing the current "compute-the-JWS-then-remove-the-body" method.

This would remove the `b64` and `crit` fields from the JWS header, and would add a new sentence describing how to calculate the body hash, dropping the dependency on RFC7797:

> The payload of the JWS object is the SHA-256 hash of the serialized body of the request

This significantly cleans up the client code, since previously the client would have to generate the JWS, the pull out the header and signature components and reassemble them with an empty body, creating a string like "XXXXX..XXXXX". Now it can use a JWS library without any additional hacks on top.

adeinega:
Such JWS header wouldn't qualify RFC7797 in this case then.

aaronpk:
how so?

adeinega:
RFC 7797 requires you to include "b64" in the "crit" header parameter and set the "b64" header parameter to `false`.

Lots of existing JWT parsers obey the "crit" parameter. As you mention, you may remove these parameters. Although, it will result in an inability to parse such JWTs by parsers that already support RFC 7797.

adeinega:
Some of the existing parsers provide a convenient way to get "XXXXX..XXXXX" without any hacks on top. For example, take a look at method `getDetachedContentCompactSerialization(...)` in https://bitbucket.org/b_c/jose4j/wiki/JWS%20Examples#markdown-header-using-the-rfc-7797-jws-unencoded-payload-option.

aaronpk:
Not relying on the additional rules specified in RFC7797 seems like a benefit, less specs to read and less edge cases to handle. Removing a dependency on another spec is a benefit.

jricher:
Right, these wouldn't be RFC7797 values anymore, they would just be plain JWS objects with a specially calculated payload. We still will have the "XXXX..XXXX" format for payload-less requests (requests without the body, such as GET and DELETE).

jricher:
I think I'm in favor of this idea, personally. It still fulfills the requirement of covering the message body with the signature method, but it does so in a way that's perhaps more native to most JOSE developers. It's also more conducive to the kinds of deployment models discussed in #196

adeinega:
[...]

PR #207, opened 2021-03-05
Mnr editorial changes for jws and jwsd

adeinega:
These changes

1. replace `A JWS signature object` with `A JWS object`
2. replace `JWS header` with `JOSE header`
3. replace `JWS field` with `JOSE parameter`

in Detached JWS and Attached JWS sections.


jricher:
I think the proposal in #201 could have some impact on this language, as well as making sure both attached and detached JWS methods are fully aligned.

adeinega:
#201 could definitely have some impact. Although, these changes touch only the used terminology.

PR #208, opened 2021-03-08
Minor changes on the language in section "Presenting Access Tokens".

adeinega:


jricher (changes requested):
Good catches, and they should go in, but the next paragraph is also going to need some edits.

https://www.ietf.org/archive/id/draft-ietf-gnap-core-protocol-04.html#section-7.2-6

The logic is now:

 - if `bound` == `false` => bearer
 - if `bound` == `true` && `key` is absent => bind to client instance's presented key and proof
 - if `bound` == `true` && `key` is present => bind to the value in `key`
 - if `bound` == `false` && `key` is present => error
 
 Previously, this was syntactically impossible as the `key` field's value folded in all of this functionality, and this PR is cleaning up some of the fallout from that change in syntax.

adeinega:
It may make sense to change the order to something like

1. `bound` == `true` && `key` is present
2. `bound` == `true` && `key` is absent
3. `bound` == `false`
4. `bound` == `false` && `key` is present

jricher (approved):
Thank you, this is great! 

Aligns with the language in #130

adeinega:
I did changes for https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/208#issuecomment-793485592.

There are lots of other things that, I believe, could be improved & changed for the "Presenting Access Tokens" section and they will be addressed in separate PRs in the future.

PR #209, opened 2021-03-08
These are examples of not sender-constrained tokens. Hence, the "bound" v…

adeinega:
…alue should be the boolean false.

Please, correct me if I wrong.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1608,8 +1608,7 @@ a [callback nonce](#response-interact-callback), and a [continuation response](#
     },
     "continue": {
         "access_token": {
-            "value": "80UPRY5NM33OMUKMKSKU",
-            "bound": true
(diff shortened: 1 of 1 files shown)


jricher (changes requested):
The intent was to make the examples bound tokens by default, so instead of changing the examples, any text surrounding these examples referring them as bearer needs to be adjusted appropriately.

adeinega:
Ok, I will take care of it a bit later. I also suggest omitting `"bound": true` from other examples as it's a default value.

adeinega:
Done.

jricher:
We'll need to add an example of `bound: "true"` in at least one location, but otherwise these changes are good.

adeinega:
Done in https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/209/commits/23bd329e9c764cedfe6a8dd1e26b557aa8fda9a6.

PR #211, opened 2021-03-11
updated diagrams, included subject-only flow

jricher:


draft-ietf-gnap-core-protocol.md
```diff
@@ -653,24 +702,82 @@ expired access token at the AS using the token's management URL.
...
     a new access token and can also include updated token management 
     information, which the client instance will store in place of the values 
     returned in (2).
-   
+
+### Requesting User Information {#sequence-user}
+
+In this scenario, the client instance is not calling an RS, and does not
+request an access token. Instead, the client instance only requests
+and is returned [direct subject information](#response-subject). Many different
```
fimbault, line 730: maybe explain why we say "direct".

fimbault (approved):
That could be useful indeed.
A few comments:
- not sure if that's really necessary, but since it's a fairly different scenario, maybe that could even be an additional response example in section 3 (and there we could use email, while elsewhere we'd use opaque).
- I checked our definition for the AS and we might just need to change very slightly (and -> or?): "server that grants delegated privileges to a particular instance of client software in the form of an access token **or** other information (such as subject information)." Because "and" means we need an access token.

draft-ietf-gnap-core-protocol.md
jricher: I was figuring that this would be updated when we do a scrub of the "subject information" terminology. For now I wanted to keep it consistent with the rest of the document.

jricher:
@fimbault Great comments.

 - I'll add a secondary example to section 3
 - I'll change that "and" to "or" -- even though it's not normative it's a good idea to give the right impression

draft-ietf-gnap-core-protocol.md
```diff
@@ -653,24 +702,82 @@ expired access token at the AS using the token's management URL.
...
-6. The AS validates the rotation request including the signature
+8. The AS validates the rotation request including the signature
     and keys presented in (5) and returns a 
     [new access token](#response-token-single). The response includes
     a new access token and can also include updated token management 
     information, which the client instance will store in place of the values 
     returned in (2).
-   
+
+### Requesting User Information {#sequence-user}
```
adeinega, line 726: It may make sense to mention that the AS authorization server MAY limit the amount of information returned in this flow.

draft-ietf-gnap-core-protocol.md
```diff
@@ -601,22 +636,32 @@ without the need for a RO to be involved at runtime to approve the decision.
...
 1. The client instance [requests access to the resource](#request). The client instance does not
     send any interactions modes to the server.
 
 2. The AS determines that the request is been authorized, 
     the AS grants access to the information
-    in the form of [access tokens](#response-token) and
-    [direct subject information](#response-subject) to the client instance.
+    in the form of [access tokens](#response-token) to the client instance.
+    Note that [direct subject information](#response-subject) is not
+    generally applicable in this use case, as their is no user involved.
```
adeinega, line 659: `there is no user`...

draft-ietf-gnap-core-protocol.md
```diff
@@ -601,22 +636,32 @@ without the need for a RO to be involved at runtime to approve the decision.
...
+
+3. The client instance [uses the access token](#use-access-token) to call the RS.
+
+4. The RS validates the token and returns an appropriate response for the
```
adeinega, line 663: I would suggest using `the access token` instead of `the token` for clarity.

draft-ietf-gnap-core-protocol.md
```diff
@@ -420,6 +430,11 @@ that returns from the interaction.
     in the form of [access tokens](#response-token) and
     [direct subject information](#response-subject) to the client instance.
 
+10. The client instance [uses the access token](#use-access-token) to call the RS.
```
adeinega, line 433: For the sake of consistency, I would suggest using only "presents the access token" or "uses the access token" everywhere.

draft-ietf-gnap-core-protocol.md
```diff
@@ -237,21 +237,24 @@ the information needed at a given stage is communicated out of band
...
-            +                            +      
-            +                            +      
-           (A)                          (B)       
-            +                            +        
-            +                            +         
-        +--------+                       +       +------------+
-        | Client |--------------(1)------+------>|  Resource  |
-        |Instance|                       +       |   Server   |
+protocol flow. Additionally some components may not be involved
+in all use cases. For example, a client instance could be calling the
```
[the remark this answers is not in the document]
jricher: Here it's appropriate to be talking about a single instance of a client acting in the context of a single request.

draft-ietf-gnap-core-protocol.md
jricher: Thank you, I actually saw this during editing but missed it here!

draft-ietf-gnap-core-protocol.md
jricher: These are examples and so adding a normative requirement isn't appropriate, but I can add a note in the process list. Any normative requirements should be in the section on subject information (response-subject).

draft-ietf-gnap-core-protocol.md
```diff
@@ -369,6 +374,11 @@ that returns from the interaction.
     |        |                                  |        |
     |        |<-(9)----- Grant Access ----------|        |
     |        |                                  |        |
+    |        |                                  |        |     +--------+
+    |        |--(10)-- Access API ---------------------------->|   RS   |
```
adeinega, line 378: My apologies if it's a bit unrelated to the proposed changes but IMO, a client instance doesn't necessarily access some APIs on an RS. It makes requests to the protected resource. Hence... does `GET /protected.docx HTTP/1.1` fall into `Access API`?

Section 1.2 (Roles) defines a client as

> an application operated by an end-user that consumes resources from one or several RSs

jricher: The protected resource is an API hosted by the resource server, RS. All of this is inline with each other, but the precision of the definitions can of course be refined!

jricher:
@fimbault @adeinega Thanks for the feedback, I've addressed the issues raised.

draft-ietf-gnap-core-protocol.md
```diff
@@ -653,24 +704,87 @@ expired access token at the AS using the token's management URL.
...
     and keys presented in (5) and returns a 
     [new access token](#response-token-single). The response includes
     a new access token and can also include updated token management 
     information, which the client instance will store in place of the values 
     returned in (2).
-   
+
+### Requesting User Information {#sequence-user}
+
+In this scenario, the client instance is not calling an RS, and does not
```
adeinega, line 731: Will
`In this scenario, the client instance neither calls an RS, nor requests an access token.`
or
`In this scenario, the client instance is neither calling an RS, nor requesting an access token.`
or 
`In this scenario, the client instance does not call an RS and does not request an access token. `
work better?

jricher: You're right, that is better. Thanks!

draft-ietf-gnap-core-protocol.md
```diff
@@ -237,21 +237,24 @@ the information needed at a given stage is communicated out of band
...
-        +------------+             +------------+
-            +                            +      
-            +                            +      
-           (A)                          (B)       
-            +                            +        
-            +                            +         
-        +--------+                       +       +------------+
-        | Client |--------------(1)------+------>|  Resource  |
-        |Instance|                       +       |   Server   |
+protocol flow. Additionally some components may not be involved
```
adeinega, line 240: A few more minor editorial suggestions...

The first is `Additionally, some components`. Then, the previous sentences already use `For example` a couple of times and you may want to replace it with `To illustrate` or something similar.

Lastly, I suggest using `a client instance can call the AS to get the direct user information and does not need to get an access token to call an RS.`.

jricher:
See also #217 for additional diagrams

PR #220, opened 2021-03-18
change subject type to format to comply with secevent draft07

fimbault:
align with [draft-07](https://datatracker.ietf.org/doc/html/draft-ietf-secevent-subject-identifiers-07)

I think however this might be a problem when we already use `format` in the text, in many places (token format, key format, assertion format etc.). Ex in section 2.2, we would have formats (relates to subject identifiers) and just below we discuss assertions: "assertions (array of strings)  An array of requested assertion formats." Of course we could change when there's potential confusion, but it shows it's not that straightforward. 

Therefore I submitted this PR to see what it does, but I think we should hold that change and ask why format is the best term for the job. Types seems more adequate a priori (a type of identifier), but there's probably a reason why they chose format instead. Even if it's still a draft, we should also ask whether it's considered stable, since our syntax depends on it. Of course, if it changes later, a mapping is possible, but that's extra gruntwork for all parties (one of the value of using sub_ids is to make the mapping straighforward). 

Please let me know what you think. If agreed, I can reach out to the editors.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -760,7 +760,7 @@ A non-normative example of a grant request is below:
     },
     "capabilities": ["ext1", "ext2"],
     "subject": {
-        "subject_types": ["iss_sub", "email"],
+        "formats": ["iss_sub", "email"],
         "assertions": ["id_token"]
     }
 }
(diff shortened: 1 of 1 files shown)


tplooker:
This PR is also related to #221 as we must consider when more information about a particular subject identifier format needs to be communicated.

jricher (approved):

Issue #225, opened 2021-03-29
Explicit typing for jws and jwsd key proofing mechanisms


adeinega:
The idea is to minimize ways to reuse JWS objects [for other purposes](https://tools.ietf.org/html/rfc8725#section-3.11). These attacks are sometimes referred as _Swapping JWTs_ and _Cross-JWT Confusion_). 

See also discussion in https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/202.

jricher:
Closed by #226

PR #226, opened 2021-03-29
The "typ" parameter for jws and jwsd key proofing mechanisms.

adeinega:
Suggested changes for https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/225.


fimbault:
This PR is now pending merge, linking with a new issue for verification #231 (to check the value based on typ).

PR #227, opened 2021-03-29
Replace parameter "ts" by "iat" for jws and jwsd key proofing mechanisms.

adeinega:
Suggested changes for https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/202#issuecomment-808941552

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3378,8 +3378,8 @@ htm (string)
 htu (string)
 : The HTTP URI used for this request, including all path and query components.
 
-ts (integer)
-: A timestamp of the request in integer seconds
+iat (integer)
+: A timestamp at which the JWS object was created.
 
 at_hash (string)
 : When a request is bound to an access token, the access token hash value. Its value is the 
@@ -3471,8 +3471,8 @@ htm (string)
 htu (string)
 : The HTTP URI used for this request, including all path and query components.
 
-ts (integer)
-: A timestamp of the request in integer seconds
+iat (integer)
+: A timestamp at which the JWS object was created.
 
 at_hash (string)
 : When a request is bound to an access token, the access token hash value. Its value is the 
@@ -3541,7 +3541,7 @@ This example's JWS header decodes to:
   "kid": "KAgNpWbRyy9Mf2rikl498LThMrvkbZWHVSQOBC4VHU4",
   "htm": "POST",
   "htu": "https://server.example.com/tx",
-  "ts": 1603800783
+  "iat": 1603800783
 }
 ~~~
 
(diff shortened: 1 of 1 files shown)


fimbault:
Yes, although that lacks information on the reference used. 

Either we could refer to https://tools.ietf.org/html/rfc7519#section-2 (NumericDate) or explain that iat's canonicalized value is an Integer String containing the signature's Creation Time expressed as the number of seconds since the Epoch, as defined in Section 4.16 of [POSIX.1].

yaronf:
> Yes, although that lacks information on the format (integer seconds) and reference used.
> 
> Either we could refer to https://tools.ietf.org/html/rfc7519#section-2 (NumericDate) or explain that iat's canonicalized value is an Integer String containing the signature's Creation Time expressed as the number of seconds since the Epoch, as defined in Section 4.16 of [POSIX.1].

IMO the reference to 7519 is most appropriate.

adeinega:
Done. I really like `NumericDate` but didn't want to add it initially as that would add an additional dependency to RFC 7519. There is absolutely nothing wrong with it except one thing, this RFC talks about JWTs, not JWS objects and it uses `NumericDate` for claims such as `iat` and `nbf`, nor header parameters.

adeinega:
Note, I used `number` as the JSON data type.

jricher:
Note that these are JWS header values, not JWT payload claims. I would also prefer not to pull JWT into this, since the proofs are not the same as tokens and they shouldn't be able to be confused as one. Not saying I'm against this change of syntax, but we'd be defining (and registering) a new header field anyway. 

As such, I have a slight preference to keep `ts` since this proof isn't "issued" like a token is, so much as just "created".

fimbault:
I agree keeping jwt out of this is best, hence the suggestion to simply describe what is expected (and we currently have a description that it is expressed in seconds, which goes in that direction).

fimbault:
We should consider renaming the field, but it should be distinct from iat because it's not a JWT.

adeinega:
For me, it isn't a big issue with reusing `iat` as is even if it overlaps with JWTs. We just need to remove a reference to RFC 7519 and provide our own description. The reason is that in programming languages, we offen use the same variable names like `i` below

```
String[] fruits = {"apple", "banana", pear"};
for (String i : fruits) {
    eat(i)
}

String[] veggies = {"broccoli", "carrot", "cucumber"};
for (String i : veggies) {
    eat(i)
}
```

and have zero issues with that, right?

If you want to use something distinct, I would suggest `rat` (requested at)... and yes, naming is hard.

jricher:
I would personally be more comfortable with `created` or the like, as that's what's being indicated here. It's a signature object, not a token -- it was not "requested" and it was not "issued" like a token would be.

fimbault:
so we need to choose between `rat` and `cat` ;-)

fimbault:
Closing. (Editors hat off) I suggest creating a new PR for created at, to replace ts.

fimbault:
Closing. (Editors hat off) I suggest creating a new PR for created at, to replace ts.

Issue #198, opened 2021-02-18
Definition of a Subject in section 1.3


Denisthemalice:
In section 1.3, subject is defined as:

    Subject 
    person, organization or device.

Given the current definition of a Right (ability given to a subject to perform a given operation on a resource under the control of a RS), it does not make sense to consider that a subject may also be an organization or a device. 

Delete "organization or a device".

fimbault:
We're mostly interested in having definitions which are consistent throughout the document. In the context of a right associated to an end-user, it is usually applicable to a person. But even then the right could derive from an automatic RBAC system as well. There are other places where subject is applicable to an organization or an IT system, especially for the resource owner. 
Notice the OR statement, depending on the context it may vary.

Denisthemalice:
A resource owner is still a person: it is neither an organization nor a device: it still does not make sense to consider that a subject may also be an organization or a device.

fimbault:
Not always, the resource owner can very well be your company for instance, and the consent might not require a real person (that's the all point of LDAP for instance).

Denisthemalice:
Do you mean that a subject is either an end-user or a resource owner which may then be either  a person or an organization ?
Would you have some arguments for a "subject" being a "device" ?

fimbault:
An end user is only a natural person. A RO could be a natural person, an organisation or a device. In the last 2 cases, you have a rule engine instead of a person.
An example of a device as a resource owner could be a system (ex a robot, a car, etc.) that controls specific resources. Not very common today tbh, but IoT scenarii make that likely in the future (at least I don't see a reason why it shouldn't be included in a generic definition).

Denisthemalice:
This topic is closely related to the issue #197.

 An AS manages end-users and maintains relationships with some ROs, but does not manage organizations, nor devices.

"Requesting Subject information" does not really make sense.

fimbault:
We never said the AS managed organisations or devices, and we didn't say either that it managed users. It manages access tokens.
It's not because the definition might in some cases allow additional cases that it doesn't make sense. We just need to be very specific as to what kind of subject is concerned when it doesn't apply to all.

Denisthemalice:
Humm ! An AS needs to manage end-users because an end-user needs first to authenticate to the AS before the client it is using may obtain an access token. The authentication mechanism being used is left out of the scope of the document, but an authentication exchange must exist. 

[...]

Issue #75, opened 2020-11-13
Scope of subject identifiers


jricher:
§3.4 Returning User Information: Editor's note: 

Subject identifiers here are naturally scoped to the AS; even though using an external identifier like an email address or phone number implies a global namespace in use, the association of that identifier to the current user is still under the view of the AS. Would changing the name to "as_sub_ids" or "local_sub_ids" help convey that point? Would it also be desirable to have an identifier that's globally unique by design? The "iss_sub" type almost gets us there by explicitly calling out the issuer URL, but tuples are hard to deal with in practice and so tend to get ignored in practice in the OIDC space.

fimbault:
Having reflected on that, I believe sub_ids should be scoped to the AS (but I don't see the value of changing the label). There's no way GNAP can impose a global identifier even if that existed, and there's even value for GNAP to be able to provide portability between identity systems #171

PR #229, opened 2021-03-29
update subject info definition according to sub_ids

fimbault:
Reflect the various scopes possible in sub_ids #75 #198
Some are local to the AS, some enable dereferencing


yaronf:
I'm confused. This reflects that we want to support both locally-scoped subjects and also non-locally ("federated"?) scoped ones. So does the comment in #75 with `iss_sub` subject identifiers. But the move to `opaque` means that we expect the AS to identify a subject using a single string. Why are we getting rid of the issuer URI?

fimbault:
We're not getting rid of iss_sub, which is also a valid sub_id.

jricher:
I don't think I agree with this change -- even in the case of an AS sending a globally-scoped identifier like iss/sub (or email, or phone, etc), the _connection_ between the user and that identifier is still being asserted locally by the AS. After all, nothing stops a rogue AS from assering an iss/sub pair that it is not the issuer for. And in the case of a user logging in to the AS with a federated identity, that might actually be desirable to know. 

The intent of the previous language was to say that the AS is only its own authority and not able to make truly global trusted statements. I think that this can be said better but I don't think removing "locally" gets at that and leaves it too open for misinterpretation.

fimbault:
The rationale is that "asserted by the AS" already provides that information, while the scope of the identifier is defined by the choice of sub_id.

agropper:
I'm always confused by the meaning of subject identifiers in authorization
protocols, not just GNAP. I wonder if this
https://mailarchive.ietf.org/arch/msg/txauth/ytSC3a_bw6LicwjU2YaBZpkS7z4/
framing
would be useful?

fimbault:
Add some text in the spec saying that the assertion is made locally by the AS

jricher (approved):

draft-ietf-gnap-core-protocol.md
```diff
@@ -2256,7 +2256,7 @@ Subject identifiers returned by the AS SHOULD uniquely identify the RO at the
 AS. Some forms of subject identifier are opaque to the client instance (such as the subject of an 
 issuer and subject pair), while others forms (such as email address and phone number) are
```
yaronf, line 2257: while others -> while other

draft-ietf-gnap-core-protocol.md
```diff
@@ -2256,7 +2256,7 @@ Subject identifiers returned by the AS SHOULD uniquely identify the RO at the
...
 intended to allow the client instance to correlate the identifier with other account information
-at the client instance. The client instance MUST NOT request or use any returned subject identifiers for communication
+at the client instance. The AS MUST validate the correspondance between the user and the returned subject identifiers. The client instance MUST NOT request or use any returned subject identifiers for communication
```
yaronf, line 2259: correspondance - typo

"The AS MUST validate" appears to be an important security check, but it's not clear to me what we're expecting the AS to do. It seems to me that the AS, once it has authenticated the RO (a.k.a. end user) simply returns all it knows about the RO. So there's nothing to validate.

yaronf (changes requested):

draft-ietf-gnap-core-protocol.md
jricher: @yaronf the process you're describing is one way to do the validation in question (arguably the most common way).

yaronf: In that case I would say the proposed language is confusing. How about:

> The AS MUST ensure that the returned subject identifiers only apply to the authenticated end user.

jricher, line 2259:
```suggestion
at the client instance. The AS MUST ensure that the returned subject identifiers only apply to the authenticated end user. The client instance MUST NOT request or use any returned subject identifiers for communication
```

jricher: Accepted Yaron's suggested text.

Issue #222, opened 2021-03-23
paramaters x5c and x5t#S256 and GNAP key formats


adeinega:
[RFC 7517 (JSON Web Key)](https://tools.ietf.org/html/rfc7517) defines a JWK as

> A JSON object that represents a cryptographic key.  The members of the object represent properties of the key, including its value.

The current draft does it as

> Value of the public key as a JSON Web Key [RFC7517].

that could be a minor editorial change. Although, at the same time, RFC 7517 defines parameters such as `x5c` and `x5t#S256` which, for the most part, intersect with `cert` and `cert#S256` key formats introduced in section "Key Formats". The question I have is what are the reasons to define these formats if the same goals are achievable with JWKs only?

Thus, we would use something like

```
{
  "key": {
    "proof": "jwsd",
    "jwk": {
      "kty": "RSA",
      "e": "AQAB",
      "kid": "xyz-1",
      "alg": "RS256",
      "n": "kOB5rR4Jv0GMeLaY6_It_r3ORwdf8ci_JtffXyaSx8xY...",
      "x5t#S256": "qwerty",
      "x5c": [
        "end-entity cert (base64)",
        "intermediate cert (base64)",
        "root cert(base64)"
      ]
    }
}
```

instead of

```
{
  "key": {
    "proof": "jwsd",
    "jwk": {
      "kty": "RSA",
      "e": "AQAB",
      "kid": "xyz-1",
      "alg": "RS256",
      "n": "kOB5rR4Jv0GMeLaY6_It_r3ORwdf8ci_JtffXyaSx8xY..."
    },
    "cert": "MIIEHDCCAwSgAwIBAgIBATANBgkqhkiG9w0BAQsFA..."
  }
}
```

That would allows to omit `cert` and `cert#S256` and express the same (single) key in multiple formats. Furthermore, this could allow to have a bit more simplified approach for https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/105.

fimbault:
Currently in section 7.1, the non-normative example shows a single key presented in multiple formats (PEM and JWK), but " If a key is sent in multiple formats, all the key format values MUST be equivalent" could be interpreted in different ways. Probably to be clarified.

So I guess there's different things to consider:
[...]

Issue #105, opened 2020-11-13
Rotation of bound client keys


jricher:
§8 Binding Keys:  Editor's note: 

We are going to need a way for a client to rotate its keys securely, even while an ongoing grant is in effect.

adeinega:
The easiest way for a client to rotate its keys is to handle a URL in the `jku` claim to the client's JWKS endpoint on the AS side (the same also applies to the "x5u" claim for certificates). "OpenID Connect Dynamic Client Registration" and "OAuth 2.0 Dynamic Client Registration Protocol (RFC 7591)" use the `jwks_uri` parameter for that. Personally, I don't think this approach will work that well for all possible scenarios because we can't assume that all clients have a way to publish and make sure their JWKs endpoints are always reachable for the AS. Furthermore, a client may not have or simply lost control of the domain and an attacker may make use of it. I guess it's totally fine to use this approach in one security domain where you have full control over all components and trust them (clients, the AS, DNS servers) but IMO, it isn't going to be a secure enough approach if you don't have that. That also won't work for the `cert` and `cert#S256` key formats defined in the current specification. I was never been a big fan of this approach (or I'm missing something important) but it may work even OTB if the specification doesn't forbid us to do that and a JWS library of your choice supports this.

Another option would be to introduce the client key management URI (similar to Token Management URI) which accepts an object with similar properties to what ["DPoP proof JWT" has](https://tools.ietf.org/html/draft-ietf-oauth-dpop-02#section-4.2) and where the goal of the AS to make sure that that the client has key material.

Upsides
1. may work for all currently defined, as well as for future key formats pretty much in the same fashion
2. no need to advertise the client key management URI using the Discovery mechanism which is great and inlined with "the protocol minimizes the need for any pre-flight discovery"
3. the client does not need to have any special knowledge on how to obtain and use tokens to rotate its keys
4. each client may have its own client key management URI

These are my thoughts.

fimbault:
Another item that won't be easy to handle is 7.1.1. Since ""dereferencing this value are out of scope for this specification", so is rotation.

adeinega:
I may also add yet another item for section "10.1. Introspecting a Token"

> The AS responds with a data structure describing the token's current state and any information the RS would need to validate the token's presentation, such as its intended proofing mechanism and key material

jricher:
[...]

Issue #47, opened 2020-11-13
Fetchable keys


jricher:
§2.3.2 Identifying the RC Key: Editor's note: 

We will eventually want to have fetchable keys, I would guess. Things like DID for key identification are going to be important.

jricher:
Fetchable keys can be addressed in a future feature if needed by the community.

PR #232, opened 2021-04-01
JWK definition

adeinega:
Suggested changes after the discussion in https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/222.

I believe GNAP may also require using the JWK thumbprint as a "kid" (Key ID) value as described in [RFC7638](https://tools.ietf.org/html/rfc7638) with the hashing algorithm agreed in advance, for example, SHA256. That will be a collision-resistant value and protection against misuse.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1891,8 +1891,7 @@ expires_in (integer)
 key (object / string)
 : OPTIONAL. The key that the token is bound to, if different from the
     client instance's presented key. The key MUST be an object or string in a format
-    described in {{key-format}}, describing a public key to which the
-    client instance can use the associated private key. The client instance MUST be able to
+    described in {{key-format}}. The client instance MUST be able to
     dereference or process the key information in order to be able
     to sign the request.
 
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3213,9 +3213,10 @@ proof (string)
     {{binding-keys}}. The `proof` field is REQUIRED.
 
 jwk (object)
-: Value of the public key as a JSON Web Key {{RFC7517}}. The object MUST
-            contain an "alg" field which is used to validate the signature.
-            The object MUST contain the "kid" field to identify the key.
+: The public key and its properties represented as a JSON Web Key {{RFC7517}}.
+    A JWK MUST contain the "alg" (Algorithm) and "kid" (Key ID) parameters. The `alg`
+    parameter MUST NOT be `none`.  The JWK x5c (X.509 Certificate Chain) parameter MAY
+    be used to provide X.509 representation of the provided public key.
 
 cert (string)
 : PEM serialized value of the certificate used to
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3215,7 +3215,7 @@ proof (string)
 jwk (object)
 : The public key and its properties represented as a JSON Web Key {{RFC7517}}.
     A JWK MUST contain the "alg" (Algorithm) and "kid" (Key ID) parameters.  The `alg`
-    parameter MUST NOT be `none`.  The JWK x5c (X.509 Certificate Chain) parameter MAY
+    parameter MUST NOT be `none`.  The "x5c" (X.509 Certificate Chain) parameter MAY
     be used to provide X.509 representation of the provided public key.
 
 cert (string)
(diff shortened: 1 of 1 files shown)


fimbault:
+1 for "The alg parameter MUST NOT be none", and the rest of the proposed change seems fine to me.

adeinega:
GNAP may also allow omitting the private key and specify "jku" (JWK set URL) and "x5u" (X.509 URL) parameters instead. That will allow rotation of the keys (only for the jwk key format) by the client at its jwks_uri endpoint. See also https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/105#issuecomment-806153772.

In this particular case, I'm guessing it may not allow mixing different parameters such as the public key and `jku` or both `jku` and `x5u`. Although, it's debatable.

If allowed, the AS MUST NOT blindly follow these parameters per recommendations 
 at https://tools.ietf.org/html/rfc8725#legend:~:text=blindly%20following%20a%20%22.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3213,9 +3213,10 @@ proof (string)
     {{binding-keys}}. The `proof` field is REQUIRED.
 
 jwk (object)
-: Value of the public key as a JSON Web Key {{RFC7517}}. The object MUST
-            contain an "alg" field which is used to validate the signature.
-            The object MUST contain the "kid" field to identify the key.
+: The public key and its properties represented as a JSON Web Key {{RFC7517}}.
+    A JWK MUST contain the "alg" (Algorithm) and "kid" (Key ID) parameters.  The `alg`
+    parameter MUST NOT be `none`.  The "x5c" (X.509 Certificate Chain) parameter MAY
+    be used to provide X.509 representation of the provided public key.
```
skreborn, line 3219: For consistency's sake, enclose all variables name in `backticks` and literal string values in "quotes".

```suggestion
: The public key and its properties represented as a JSON Web Key {{RFC7517}}.
    A JWK MUST contain the `alg` (Algorithm) and `kid` (Key ID) parameters. The `alg`
    parameter MUST NOT be "none". The `x5c` (X.509 Certificate Chain) parameter MAY
    be used to provide the X.509 representation of the provided public key.
```

skreborn (changes requested):

draft-ietf-gnap-core-protocol.md
adeinega: I would leave two spaces after a period as it was before.

jricher: A style note from the editors: There should only be one space before period inside of text. Any existing text that uses two spaces before a dot will eventually get cleaned up. It's a minor detail, but if there's a chance to fix or avoid it in a PR, we'll try to do so.

jricher:
Fetchable keys were under consideration in #47 but the WG decided to table that for now and focus on keys by value and by direct reference instead, due to the complexities around security of key fetching. We'll still need to deal with rotation of keys by value even if we allow fetchable keys, and the rotation of fetched keys isn't straightforward are susceptible to cache poisoning and other attacks.

In this PR, I agree with @fimbault that the 'must not be `none`' text is good, but the `x5c` text might not be helpful because of what's discussed in #222

adeinega:
https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/222 suggested replacing `cert` and `cert#S256` with JWKs and their properties. It isn't quite the same as the changes here
Yes, we can remove

> The `x5c` (X.509 Certificate Chain) parameter MAY be used to provide the X.509 representation of the provided public key.

that allows specifying the `x5c` parameter _only corresponding_ to the provided public key if WG wants to do so.

Note, JWKs remain to be very flexible and if the goal is to allow only public keys we probably should explicitly specify what is allowed. Interestingly enough, a JWK as the format can contain both public and private keys and I don't think it's desirable to allow.

jricher:
The definition already states that these are public keys. Shared keys aren't allowed to be passed by value, but could instead be passed by reference through either the client instance identifier or the key reference, if either is applicable to the situation. 

I'd like to hear feedback from other implementors on the utility of including the language around `x5c` in this definition here.

fimbault:
Related to public/private keys:
- The text in section 7 is clear I think
- We might want to remove the part in bold "The key MUST be an object or string in a format described in Section 3.2.1, describing a public key **to which the client instance can use the associated private key.**" I don't think that brings anything.

adeinega:
The current version of https://bitbucket.org/openid/fapi/src/master/Financial_API_Simple_HTTP_Message_Integrity_Protocol.md allows specifying `x5c`, as well as `x5u` and `jwks_uri` for a DPoP proof. The PR doesn't say anything about `x5u` and `jku` / `jwks_uri`. Yes... it could be a separate issue.

I'm guessing that certificates are only one way to go in certain depolyments.

Issue #114, opened 2020-11-13
Resource Server Section


jricher:
§10 Resource Servers: Editor's note: 

This section is for discussion of possible advanced
functionality. It seems like it should be a separate document or set of
documents, and it's not even close to being well-baked. This also adds
additional endpoints to the AS, as this is separate from the token
request process, and therefore would require RS-facing discovery or
configuration information to make it work. Also-also, it does presume
the RS can sign requests in the same way that a client does, but
hopefully we can be more consistent with this than RFC7662 was able to
do.

Denisthemalice:
Section 10 mentions:

> In some deployments, a resource server will need to be able to call the AS for a number of functions.

Calling back the AS, allows the AS to know to the existence of the RS. There are cases where the AS should be prevented to know which the RS is.

When doing "Token Introspection" (Section 10.1), the AS may communicate to the RS some privileges without the client (and hence the end-user) knowing it (and thus **_without the end-user consent_**). 

"Deriving a downstream token" (Section 10.2) is described to allow a RS to act as a client instance and request a derived access token for a secondary RS. This can be done without a recall to the AS, if the name of the service to which the first RS belongs to is included into the access token. See the issue #203 which contains more details on how to do it.

"Registering a Resource Handle" (Section 10.3. ) allows the AS to know which RSs are trusting it. This should be avoided for RSs as soon as there is no bi-lateral trust relationships with the AS. This should also be avoided to allow large scale developments. 
A RS trusts some ASs to some extent, but the ASs do not necesarilly need to know which RSs are trusting them.

Requesting Resources With Insufficient Access (Section 10.4) is addressed in another issue. 
See "Requesting resources with insufficient access" #203

The whole section would need to be re-written.

jricher:
Section 10 is made of optional elements. Token introspection is not required, you could alternatively use a self-contained token. Deriving a token without the AS's involvement is deeply problematic for most use cases as the AS is the root of trust in the system and overriding that would subvert the security model. Yes, introducing the RS to the AS lets the AS know about the RS as this is the point.

Denisthemalice:
(Section 10.4) Requesting Resources With Insufficient Access is NOT an optional element.
 Nevertheless, it is addressed in another issue.  See "Requesting resources with insufficient access" #203.
 
At the moment the text states:

> In some deployments, a resource server will need to be able to call the AS for a number of functions. 

[...]

PR #242, opened 2021-04-06
Gathering consent and authorization through interaction and other means

### LINKED ISSUE #224: Expand "user interaction" discussion
The "user interaction" section of GNAP has always been intended to allow for many different kinds of practices, including things like discussed in #168 

The current writing makes it seem limited to just the RO interacting directly with the AS, and then only specified through a web browser. The section needs to be rewritten to take into account the variety of different acts that the AS can allow during negotiation process that results in the delegation to the client instance. 

See also #72

jricher:
This rewrite changes the framing of the "interaction" sections to showcase that the interactions themselves don't have to take place through web pages hosted at the AS. The AS in fact has a wide variety of methods that it can use, with interaction (web based or otherwise) being among the most common and flexible. The AS could also use elements like digital wallets, shared secure storage, messaging fabrice, or bound assertions to communicate. 

Furthermore, the AS itself could be deployed in a variety of ways, with the interactive components being separate from the core delegation functions of the AS. 

Closes #224

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -347,7 +347,8 @@ In this example flow, the client instance is a web application that wants access
 of the current user, who acts as both the end-user and the resource
 owner (RO). Since the client instance is capable of directing the user to an arbitrary URL and 
 receiving responses from the user's browser, interaction here is handled through
-front-channel redirects using the user's browser. The client instance uses a persistent session
+front-channel redirects using the user's browser. The interaction URL is functionally
+a direct service of the AS in this example. The client instance uses a persistent session
 with the user to ensure the same user that is starting the interaction is the user 
 that returns from the interaction.
 
@@ -442,7 +443,8 @@ An example set of protocol messages for this method can be found in {{example-au
 
 In this example flow, the client instance is a device that is capable of presenting a short,
 human-readable code to the user and directing the user to enter that code at
-a known URL. The client instance is not capable of presenting an arbitrary URL to the user, 
+a known URL. The URL the user enters the code at is a functional component of the
+AS in this example. The client instance is not capable of presenting an arbitrary URL to the user, 
 nor is it capable of accepting incoming HTTP requests from the user's browser.
 The client instance polls the AS while it is waiting for the RO to authorize the request.
 The user's interaction is assumed to occur on a secondary device. In this example
@@ -733,7 +735,7 @@ In this scenario, the client instance does not call an RS and does not
 request an access token. Instead, the client instance only requests
 and is returned [direct subject information](#response-subject). Many different
 interaction modes can be used in this scenario, so these are shown only in 
-the abstract here.
+the abstract as functions of the AS here.
 
 ~~~
     +--------+                                  +--------+         +------+
@@ -803,7 +805,7 @@ client (object / string)
 : Describes the client instance that is making this request, including 
     the key that the client instance will use to protect this request and any continuation
     requests at the AS and any user-facing information about the client instance used in 
-    interactions at the AS. {{request-client}}
+    interactions. {{request-client}}
 
 user (object / string)
 : Identifies the end-user to the AS in a manner that the AS can verify, either directly or
@@ -1348,16 +1350,18 @@ return an error.
 
 ## Interacting with the User {#request-interact}
 
-Many times, the AS will require interaction with the RO in order to
+Often, the AS will require [interaction with the RO](#authorization) in order to
 approve a requested delegation to the client instance for both access to resources and direct
 subject information. Many times the end-user using the client instance is the same person as
-the RO, and the client instance can directly drive interaction with the AS by redirecting
-the end-user on the same device, or by launching an application. Other times, the 
+the RO, and the client instance can directly drive interaction with the end user by facilitating
+the process through means such as redirection to a URL or launching an application. Other times, the 
 client instance can provide information to start the RO's interaction on a secondary
 device, or the client instance will wait for the RO to approve the request asynchronously.
-The client instance could also be signaled that interaction has completed by the AS making
-callbacks. To facilitate all of these modes, the client instance declares the means that it 
-can interact using the `interact` field. 
+The client instance could also be signaled that interaction has concluded through a
+callback mechanism.
+
+The client instance declares the parameters for interaction methods that it can support 
+using the `interact` field. 
 
 The `interact` field is a JSON object with three keys whose values declare how the client can initiate
 and complete the request, as well as provide hints to the AS about user preferences such as locale. 
@@ -1421,7 +1425,7 @@ This specification defines the following interaction start modes as an array of 
 
 "redirect"
 : Indicates that the client instance can direct the end-user to an arbitrary URL
-    at the AS for interaction. {{request-interact-redirect}}
+    for interaction. {{request-interact-redirect}}
 
 "app"
 : Indicates that the client instance can launch an application on the end-user's
@@ -1429,7 +1433,7 @@ This specification defines the following interaction start modes as an array of 
 
 "user_code"
 : Indicates that the client instance can communicate a human-readable short
-    code to the end-user for use with a stable URL at the AS. {{request-interact-usercode}}
+    code to the end-user for use with a stable URL. {{request-interact-usercode}}
 
 #### Redirect to an Arbitrary URL {#request-interact-redirect}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+another party, or another piece of software through its [interaction start](#request-interact-start) request.
+In many cases, the end user is delegating their own access as RO to the client instance. Here, the 
+AS needs to determine the identity of the end user and will often need to interact directly with
+the end user to determine their status as an RO and collect their consent. If the AS has determined
+that authorization is required and the AS can support one or more of the requested interaction start
+methods, the AS returns the associated [interaction start responses](#response-interact). The client
+instance SHOULD [initiate one or more of these interaction methods](#interaction-start) in order to 
+facilitate the granting of the request. If more than one interaction start method is available,
+the means by which the client chooses which methods to follow is out of scope of this specification.
+The client instance MUST use each interaction method once at most.
```
fimbault, line 2449: Not very clear what we mean by "MUST use each interaction method once at most", and seems like a requirement better suited for the sub section related to interact

Also a bit in contradiction with what follows "If the AS and client instance have not reached a state where the delegation can be granted, the AS and client instance can repeat the interaction process as long as"

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+either in response to a signal indicating the [finish of the interaction](#interaction-finish), through
+polling, or through some other method defined by an extension of this specification.
+
+If the AS and client instance have not reached a state where the delegation can be granted, the
+AS and client instance can repeat the interaction process as long as the AS supplies the client
+instance with [continuation information](#response-continue) to facilitate the ongoing requests.
+
+## Interaction Start Methods {#interaction-start}
+
+To initiate an interaction start method indicated by the
```
fimbault, line 2461: We should probably shortly remind here what we mean by "interaction start". Notice that the link to 3.3 never talks about start (so far it's only discussed in the request, but it's only implicit in the response)

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+If the interaction component at the user code URI is not hosted by the AS directly, the means of communication between
+the AS and this URI, including communication of the user code itself, are out of scope for this specification.
 
-The user code URL MUST be reachable from the RO's browser, though
-note that the RO MAY open the URL on a separate device from the client instance
-itself. The user code URL MUST be accessible from an HTTP GET request,
-and MUST be protected by HTTPS or equivalent means.
+When the RO enters this code at the user code URI,
+the AS MUST uniquely identify the pending request that the code was associated
+with. If the AS does not recognize the entered code, the interaction component MUST
```
fimbault, line 2516: there could also be a timeout (cf 3.3: "The time in which this code will be accepted SHOULD be short lived, such as several minutes.")

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+display an error to the user. If the AS detects too many unrecognized codes
+enter attempts, the interaction component SHOULD display an error to the user.
```
fimbault, line 2518: the AS may also decide to take any appropriate action in case there are too many errors (like freeze temporarily). If there's a maximum number of attempts, it probably should be displayed to the user before that kind of decision is taken

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+the URI is usually be opened on a separate device from the client instance
+itself. The URI MUST be accessible from an HTTP GET
+request and MUST be protected by HTTPS or equivalent means.
 
+### Interaction through an Application URI {#interaction-app}
 
-## Interaction through an Application URI {#interaction-app}
+When the client instance is directed to launch an application through the
+["app"](#response-interact-app) mode, the client launches the
+URL as appropriate to the system. The means by which the AS and the
```
fimbault, line 2531: for clarity, we might want to add at least an example of such means (ex: deep linking for mobile)?

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
-available with the current request, the AS MUST follow the appropriate
-method at the end of interaction to allow the client instance to continue. If
-this mode is not available, the AS SHOULD instruct the RO to
-return to their client instance upon completion. Note that these steps
-still take place in most error cases, such as when the RO has denied
-access. This pattern allows the client instance to potentially recover from the error
-state without restarting the request from scratch by modifying its
-request or providing additional information directly to the AS.
+If an interaction ["finish"](#response-interact-finish) method is
+associated with the current request, the AS MUST follow the appropriate
```
fimbault, line 2538: caveat: we said earlier that if an error happens before, the AS MUST not follow

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
 
 The AS MUST create an interaction reference and associate that
 reference with the current interaction and the underlying pending
-request. This value MUST be sufficiently random so as not to be
+request. This interaction reference value MUST be sufficiently random so as not to be
 guessable by an attacker. The interaction reference MUST be
-one-time-use.
+one-time-use to prevent interception and replay attacks.
 
 The AS MUST calculate a hash value based on the client instance and AS nonces and the
```
fimbault, line 2549: that's a minor comment, but nonce in british english doesn't mean at all the same as in American ...

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+If the AS determines that additional runtime authorization is required, the AS can either deny the request
+outright or use a number of means at its disposal to gather that authorization from the appropriate ROs, 
+including for example:
+
+- starting interaction with the end user through the client instance
+- challenging the client instance through a response mechanism
+- request that the client instance present specific additional information
+- contacting a RO through an out-of-band mechanism
+- contacting an auxiliary software process through an out-of-band mechanism
+- any additional mechanisms defined by extensions to this protocol
```
fimbault, line 2411: I get the idea of that last bullet point, the problem however is that it's not entirely clear what is core and what is an extension, even if the bullet items just above

fimbault (commented):
Interesting and useful PR, a few comments

agropper:
I agree with the intent but I think the starting point of the first two
sequences could be improved if we were consistent in having Step (1) always
be Client Instance --- Request Access ---> AS. If anything that happens
before Step (1) is in scope for GNAP, it would be useful to state that
clearly up-front. If it's a difference between GNAP Core and Extensions,
then that could be stated up-front as well.

fimbault:
so changing diagram 1.4.1? (removing 1)

jricher:
@agropper changing that diagram will be the addressed in the PR targeted at fixing #114 (this hasn't been done/pushed yet)

Remember, each PR isn't aiming at perfect, just "better" :)

draft-ietf-gnap-core-protocol.md
```diff
@@ -2071,18 +2071,20 @@ If the client instance indicates that it can [redirect to an arbitrary URL](#req
...
 ~~~
 
-The interaction URL returned represents a function of the AS but MAY be completely
+The URL returned is a function of the AS, but the URL itself MAY be completely
 distinct from the URL the client instance uses to [request access](#request), allowing an
 AS to separate its user-interactive functionality from its back-end security
-functionality.
+functionality. If the AS does not directly host the functionality accessed through
+the given URL, then the means for the interaction functionality to communicate
+with the rest of the AS are out of scope for this specification.
```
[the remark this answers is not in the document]
jricher: Agree, but since it's out of scope here I'm not sure there's much we can really say about it.

draft-ietf-gnap-core-protocol.md
```diff
@@ -2368,108 +2370,199 @@ unknown_request
...
+defined and supported by GNAP for fulfilling these requirements, including information send by the client instance, information
+gathered through the interaction process, and information supplied by external parties. An individual AS
+can define its own policies and processes for deciding when and how to gather the necessary authorizations
+and consent.
+
+The client instance can supply information indicating that the client instance has the
+right to be granted the requested delegation immediately, including:
+
+- the identity of the client instance, known from the presented keys or associated identifiers
+- the identity of the end user presented in the `user` request parameter
```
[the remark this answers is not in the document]
jricher: Here we're specifically using "end user" as defined in the terminology section.

draft-ietf-gnap-core-protocol.md
jricher: I haven't seen an alternative embraced by the community yet, but I'd be happy for us to change if a good suggestion is made.

draft-ietf-gnap-core-protocol.md
jricher: This is tricky since for most error conditions we actually do want the AS to follow the "finish" method even in error cases. The only cases where the finish shouldn't be followed are ones where the request or the URI are fishy.

PR #243, opened 2021-04-08
Fix the payload of the introspection request.

adeinega:
Otherwise, there isn't any way for the AS to identify which client makes the introspection request (or at least, there is no obvious way for me).

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4277,6 +4277,7 @@ Detached-JWS: ejy0...
 
 {
     "access_token": "OS9M2PMHKUR64TB8N6BW7OZB8CDFONP219RP1LT0",
+    "client": "7C7C4AZ9KHRS6X63AJAO"
 }
 ~~~
 


adeinega:
Yet another (very) minor comment from me. Lots of examples in this specification use information about the client provided by the _value_. I suggest providing it by the _reference_ in order to improve readability and make these examples significantly shorter. For examples, these sections

- 7.3.1. Detached JWS
- 7.3.2. Attached JWS
- 7.3.3. Mutual TLS
- 7.3.4. Demonstration of Proof-of-Possession (DPoP)

and so forth.

fimbault:
I don't think so, at least not in this way.
The introspection is used by the RS, who doesn't know the client ref used between the client and the AS. The AS is the one that emitted the token, so it can find it. 
However, if for privacy reasons we decide to hide the RS to the AS, it would become tricky for the AS to know whether it's a legitimate ask. That's where (maybe) provenance could be useful. 

For the client, there is also a management endpoint (cf section 6). 

As for using references and reducing the size of the text, that's a good remark, although we need to make sure we do provide examples for every value.

adeinega:
@fimbault, I'm a bit confused then. `client` in this payload corresponds to the client instance on the RS side. The AS should know it in order to verify the signature for

> The RS signs the request with its own key (not the client instance's key or the token's key).

fimbault:
The key here is from the RS (as the comment describes), because it's an RS-AS communication (the client isn't involved). Except from a detached JWS instead of a basic auth, there's currently no difference to what exists in OAuth2 (RFC 7662)

adeinega:
RFC 7662 requires some form of authorization to access the introspection endpoint but leaves low-level details about them out of scope. You may use the client authentication (`client_id` + `client_secret` in the request body or the HTTP Basic auth scheme) which is extremely common... and/or something completely different.

I thought we would also have/allow the same `client` too as well.

jricher:
If we have a field here to identify the calling AS, it will not be named `client` as that's been a huge point of confusion in the OAuth introspection implementation world. The important difference from OAuth is that the key-presentation mechanisms used by the client for authorization in GNAP are abstracted so that an RS can use the same mechanisms in different calls. That parallelism is something I personally wish we'd had when working on RFC7662 and it's something I think we can do better about here. 

This whole section is being pulled out as part of #114 so we'll have a chance to address it more directly there, and so I'm inclined to close this PR and file an issue against the new document once we have that in hand.

adeinega:
Ok, I'm closing this PR then. We'll return to this topic again in https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/114 taking into account

> hopefully we can be more consistent with this than RFC7662 was able to do.

from it.

Let me know if you would like to see a separate PR for https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/243#issuecomment-815493321.

PR #246, opened 2021-04-13
Extract RS-focused protocol elements

### LINKED ISSUE #114: Resource Server Section

jricher:
This change set extracts the resource-server-focused elements of GNAP to allow the core protocol to focus on the client-AS relationship while this new document
can focus on the RS-AS relationship.

Closes #114

- remove RS section for refactoring

RS section now at: https://github.com/ietf-wg-gnap/gnap-resource-servers

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4306,6 +4306,8 @@ sure that it has the permission to do so.
     - Changed "key_proofs" to "key_proofs_supported".
     - Changed "assertions" to "assertions_supported".
     - Updated discovery and field names for subject formats.
+    - Updated discovery and field names for subject types.
+    - Refactored the RS-centric components into a new document. 
 
 - -04
     - Updated terminology.
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -43,6 +43,7 @@ normative:
            ins: R. Holz
          -
            ins: P. Saint-Andre
+    I-D.draft-ietf-gnap-resource-servers:
     RFC2119:
     RFC3230:
     RFC3986:
@@ -254,7 +255,7 @@ an access token to call an RS.
             +                         +        
             +                         +         
         +--------+                    +          +------------+
-        | Client |--------------(1)---+--------->|  Resource  |
+        | Client | (1)                +          |  Resource  |
         |Instance|                    +          |   Server   |
         |        |       +---------------+       |    (RS)    |
         |        |--(2)->| Authorization |       |            |
@@ -263,12 +264,12 @@ an access token to call an RS.
         |        |--(4)->|               |       |            |
         |        |<-(5)--|               |       |            |
         |        |--------------(6)------------->|            |
-        |        |       |               |<~(7)~~|            |
+        |        |       |               |   (7) |            |
         |        |<-------------(8)------------->|            |
         |        |--(9)->|               |       |            |
         |        |<-(10)-|               |       |            |
         |        |--------------(11)------------>|            |
-        |        |       |               |<~(12)~|            |
+        |        |       |               |  (12) |            |
         |        |-(13)->|               |       |            |
         |        |       |               |       |            |
         +--------+       +---------------+       +------------+
@@ -283,12 +284,11 @@ an access token to call an RS.
     behalf of the RO. This could identify the RS the client instance needs to call,
     the resources needed, or the RO that is needed to approve the 
     request. Note that the RO and end-user are often
-    the same entity in practice.
+    the same entity in practice, but some more dynamic processes are discussed in
+    {{I-D.draft-ietf-gnap-resource-servers}}.
     
-- (1) The client instance [can attempt to call the RS](#rs-request-without-token) to determine 
-    what access is needed.
-    The RS informs the client instance that access can be granted through the AS. Note that
-    for most situations, the client instance already knows which AS to talk to and which
+- (1) The client instance determines what access is needed and which AS to approach for access. Note that
+    for most situations, the client instance is pre-configured with which AS to talk to and which
     kinds of access it needs.
 
 - (2) The client instance [requests access at the AS](#request).
@@ -314,10 +314,9 @@ an access token to call an RS.
 - (6) The client instance [uses the access token](#use-access-token) to call the RS.
 
 - (7) The RS determines if the token is sufficient for the request by
-    examining the token, potentially [calling the AS](#introspection). Note that
-    the RS could also examine the token directly, call an internal data store,
-    execute a policy engine request, or any number of alternative methods for
-    validating the token and its fitness for the request.
+    examining the token. The means of the RS determining this access are
+    out of scope of this specification, but some options are discussed in
+    {{I-D.draft-ietf-gnap-resource-servers}}.
     
 - (8) The client instance [calls the RS](#use-access-token) using the access token
     until the RS or client instance determine that the token is no longer valid.
@@ -330,8 +329,10 @@ an access token to call an RS.
 
 - (11) The client instance [uses the new access token](#use-access-token) to call the RS.
 
-- (12) The RS determines if the new token is sufficient for the request by
-    examining the token, potentially [calling the AS](#introspection).
+- (12) The RS determines if the new token is sufficient for the request.
+    The means of the RS determining this access are
+    out of scope of this specification, but some options are discussed in
+    {{I-D.draft-ietf-gnap-resource-servers}}.
 
 - (13) The client instance [disposes of the token](#revoke-access-token) once the client instance
     has completed its access of the RS and no longer needs the token.
@@ -3186,9 +3187,8 @@ key proof together.
 - When an access token is used with no key proof, this is a bearer token request. This type of
     request is used only for calls to the RS, and only with access tokens that are
     not bound to any key as described in {{response-token-single}}.
-- When neither an access token nor key proof are used, this is an unsecured request. This 
-    type of request is used only for calls to the RS during a discovery phase as
-    described in {{rs-request-without-token}}.
+- When neither an access token nor key proof are used, this is an unsecured request. This type
+    of request is not used in the core protocol of GNAP.
 
 ## Key Formats {#key-format}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+    RFC8705:
+    RFC8693:
+    I-D.ietf-gnap-core-protocol:
+    I-D.ietf-httpbis-message-signatures:
+    I-D.ietf-oauth-signed-http-request:
+    I-D.ietf-oauth-dpop:
+    I-D.ietf-secevent-subject-identifiers:
+    OIDC:
+      title: OpenID Connect Core 1.0 incorporating errata set 1
+      target: https://openiD.net/specs/openiD-connect-core-1_0.html
```
adeinega, line 68: This link is broken.

adeinega:
Regarding 

`This specification also discusses methods for an RS to derive a downstream token for calling another chained RS...`
 
Why this specification can't state it even in a more generic way similar to what https://tools.ietf.org/html/rfc8693 does? There are use cases when we may need to exchange an access token and then use it with something completely different than chained RSs.

adeinega:
Maybe this is a bit unrelated but... The self-contained tokens usually contain one or another form of the issuer. What would the issuer for GNAP? Is it the grant endpoint?

jricher:
@adeinega I'd be happy to see proposed language to discuss! The limitation is to keep parity with the "token request" mechanism in the core draft: the goal is to get an access token, but based on the context of an existing access token. 

One of the problems I've always had with RFC8693 is that it tries to be many things at once, from access token exchange to assertion exchange to likely other use cases. If other functions should be added to this spec then we can do that, but hopefully we won't run the same risk of putting everything through the same narrow pipe. Technically since it's a basic GNAP request, you can already return subject identifiers or assertions as well, but we don't currently talk about any of that. Should we?

jricher:
@adeinega The concept of the "issuer" entirely depends on the type of token. The grant endpoint would be an obvious value for GNAP though, and if we're going to go down the route of discussing token contents from a structural standpoint (or even profile things like JWT or Biscuits or PASETO) then we can make that recommendation. However, the current text is intended to be at a much higher level than that, right now.

adeinega:
@jricher, I also wonder if GNAP is going to support encrypted token introspection responses on its core?

adeinega:
Another thing I would like to understand better if GNAP is going to rely completely on TLS... or it would

1. sign its own responses such as token introspect responses say using `Detached-JWS` (or other means)
2. provide a way to cryptographically link a request from the client to AS's response

TLS isn't going anywhere... It's a must. Although, in certain forms of envs, it's great to have an additional layer on top of it.

adeinega:
Regarding,
_However, the current text is intended to be at a much higher level than that, right now._

It's understandable. It's entirely possible I missed something in the current spec but it would be nice to say that a URL of the grant endpoint also represents something the `issuer` does in other identity-related specs.

I hope it makes some sense.

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+            },
+            "dolphin-metadata"
+        ]
+    },
+    "client": "7C7C4AZ9KHRS6X63AJAO",
+    "existing_access_token": "OS9M2PMHKUR64TB8N6BW7OZB8CDFONP219RP1LT0"
+}
+~~~
+
+The AS responds with a token for the downstream RS2 as described in
```
adeinega, line 420: I would suggest omitting the `downstream` word for https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/246#issuecomment-819066135 to make this section more generic. An RS 1 may call RS2 that belongs to a completely different sec domain. RS 2 may also trust only to its own AS 2... and there is some form of trust established between AS 1 and AS 2. The same applies to the name of this section (Deriving a downstream token). I do hope it makes sense.

jricher: Fair point, I think the current section is tightly focused on the RS going back to the same issuing AS that the original token came from -- this doesn't have to be the case, of course, but for the cases where it's not the issuing AS then as you point out there needs to be some kind of trust happening between the two AS's such that the second AS can make any sense of things. I am a little afraid of the model becoming overly abstract and not giving developers enough to hook on to, but I think it's worth pursuing. I'm going to suggest that if the WG does accept this PR then we continue that conversation in a new issue dedicated to that topic.

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+HTTP/1.1 200 OK
+Content-Type: application/json
+Cache-Control: no-store
+
+{
+    "active": true,
+    "access": [
+        "dolphin-metadata", "some other thing"
+    ],
+    "key": {
```
adeinega, line 255: What's about the rotation of keys? Shouldn't it be a list of keys? Without that, we won't have a smooth transition when the client decides to rotate its key.

I still suggest requiring the JWK thumbprint as a value for `kid` in `jwk` as protection against misuse.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3186,9 +3187,8 @@ key proof together.
 - When an access token is used with no key proof, this is a bearer token request. This type of
     request is used only for calls to the RS, and only with access tokens that are
     not bound to any key as described in {{response-token-single}}.
-- When neither an access token nor key proof are used, this is an unsecured request. This 
-    type of request is used only for calls to the RS during a discovery phase as
-    described in {{rs-request-without-token}}.
+- When neither an access token nor key proof are used, this is an unsecured request. This type
+    of request is not used in the core protocol of GNAP.
```
fimbault, line 3191: Ok, but the previous text was clearer as to what we're referring to. Could lonk to the proper section in I-D.draft-ietf-gnap-resource-servers as an example?

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+    organization: Okta
+    uri: https://aaronparecki.com
+
+  - ins: F. Imbault
+    name: Fabien Imbault
+    organization: acert.io
+    email: fabien.imbault@acert.io
+    uri: https://acert.io/
+
+normative:
```
fimbault, line 34: Probably should review that list, I guess there's a lot we don't use in here specifically.

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+resource_registration_endpoint:
+: The URL of the endpoint offering resource registration. {{rs-register-resource-handle}}
+
+grant_endpoint:
+: The grant endpoint of the GNAP AS.
+
+## Protecting RS requests to the AS {#authentication}
+
+Where noted, the RS protects its calls to the AS using any of the signature
+methods defined by GNAP. This signing method MUST cover all of the appropriate
```
fimbault, line 176: sentence to finish "all of the appropriate"

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+\[\[ [See issue #117](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/117) \]\]
+
+
+
+# Deriving a downstream token {#token-chaining}
+
+Some architectures require an RS to act as a client instance and use a derived access
+token for a secondary RS. Since the RS is not the same entity that made the initial grant
+request, the RS is not capable of referencing or modifying the existing grant. As such, 
+the RS needs to request or generated a new token access token for its use at the secondary RS.
```
fimbault, line 336: generated -> generate

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+This internal secondary token is issued in the context of the incoming access token.
+
+While it is possible to use a [token format]{#token-format} that allows for the
```
fimbault, line 339: #structure?

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+invalid access token, the RS MAY respond to the client instance with an
+authentication header indicating that GNAP needs to be used
+to access the resource. The address of the GNAP
+endpoint MUST be sent in the "as_uri" parameter. The RS MAY
+additionally return a resource reference that the client instance MAY use in
+its access token request. This
+resource reference handle SHOULD be sufficient for at least the action
+the client instance was attempting to take at the RS. The RS MAY use the [dynamic resource handle request](#rs-register-resource-handle) to register a new resource handle, or use a handle that
+has been pre-configured to represent what the AS is protecting. The
+content of this handle is opaque to the RS and the client instance.
```
fimbault, line 437: "The content of this handle is opaque to the client instance" would cover the general case (dynamic or pre-configured).

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,502 @@
...
+
+resource_registration_endpoint:
+: The URL of the endpoint offering resource registration. {{rs-register-resource-handle}}
+
+grant_endpoint:
+: The grant endpoint of the GNAP AS.
+
+## Protecting RS requests to the AS {#authentication}
+
+Where noted, the RS protects its calls to the AS using any of the signature
```
fimbault, line 175: Where noted -> I would reverse the sentence "Unless otherwise specified"

fimbault (commented):
I think that's a good starting point. 

A few general remarks:
- Seems to me that 3.3 is more an AS facing API. We could have a paragraph 3 focused on RS APIs (well known, registering a resource handle, etc.), and regroup the other items in a section that discusses how the RS reacts when a client calls it (i.e. current sections 3.3, 4 and 5)
- Will we need some kind of resource_server instance to declare a stable "resource_server": "7C7C4AZ9KHRS6X63AJAO"? (beyond tne reference, which seems chosen here by the RS?)

draft-ietf-gnap-resource-servers.md
jricher: In this design, if the key is rotated then whatever the "current" key is will be returned here, replacing the "old" key. I'm not sure I see the value in the overhead that multiple keys would have here just to address this one particular corner. To me it feels like having multiple keys tied to a single access token at any time would be problematic and lead to potential security gaps. Introspection already has to deal with cache inconsistency problems, so an RS is going to need to be robust against something like that anyway. This kind of situation can be talked about in some kind of implementation guide section, perhaps.

draft-ietf-gnap-core-protocol.md
jricher: We could point externally to that, but I took the reference out because one should never be making this kind of unprotected call in any of GNAP's core. So, this bullet was really just supposed to complete the matrix of possibilities and say "don't do this" -- the exception is in the extension that allows bootstrapping.

draft-ietf-gnap-resource-servers.md
jricher: Yes, and the tools will throw warnings for unused references that we can use to trim things -- I just left the list as-is for now as we settle on which text goes where to avoid compilation errors for now.

draft-ietf-gnap-resource-servers.md
jricher: Weird, that shouldn't have compiled with a broken link. Fixed now.

draft-ietf-gnap-resource-servers.md
jricher: correct, this is requesting resources by reference

jricher:
@fimbault :

 - Section 3.3 (introspection) is put here right now because introspection is an API presented to the RS by the AS, just like the others. I'm very open to reorganizing the whole document though as it does need to be more cohesive. This pull was a first step.
 -  This is currently covered in 3.2, the AS can pre-configure the identifiers or let the RS just present a key at runtime. This discussion needs to be expanded along with security considerations for both practices, and ways to do the binding dynamically (like return the value at runtime, the way a client instance ID is returned)

draft-ietf-gnap-resource-servers.md
adeinega: Interestingly... so, your point is that an RS should be robust to handle this. I get it. Although we may still have & deal with an extremely chatty application and there is an unavoidable delay between the decision to rotate its key by the client and what both AS and RS known.

adeinega: I still like to hear your thoughts on

`I still suggest requiring the JWK thumbprint as a value for kid in jwk as protection against misuse.`

adeinega: Regarding `with cache inconsistency problems`, to be completely pedantic, the response, as well as the spec, includes `Cache-Control: no-store`. So, it's forbidden. You may allow omitting it for the introspection endpoint's response.

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,505 @@
...
+4. The RS fulfills the request from the client instance.
+
+The RS signs the request with its own key and sends the access
+token as the body of the request.
+
+~~~
+POST /introspect HTTP/1.1
+Host: server.example.com
+Content-Type: application/json
+Detached-JWS: ejy0...
```
adeinega, line 230: `resource_server` identifier could be a part of `Detached-JWS`. In this case, you might want to omit it from the actual request. I don't see any reason to specify it twice.

draft-ietf-gnap-resource-servers.md
```diff
@@ -0,0 +1,505 @@
...
+
+{
+    "access_token": "OS9M2PMHKUR64TB8N6BW7OZB8CDFONP219RP1LT0",
```
adeinega, line 233: This section doesn't specify the required field in the introspection request. At least, the "access_token" field should be requred, right?

draft-ietf-gnap-resource-servers.md
jricher: The identifier is not defined as part of that structure, it's defined as part of the body of the request where needed. There's no need to overload it. DPoP overloads the key identifier here because it has no other choice.

draft-ietf-gnap-resource-servers.md
jricher: HTTP Cache and functional application cache are totally separate from each other. You can disable an HTTP cache but still have an application save the results and simply not make another HTTP call again.

jricher: JWK identifiers are varied and can have meaning inside of an environment (such as date/timestamps, service identifiers, etc), requiring a specific format will limit the utility of the protocol unnecessarily as it adds no additional benefits.

draft-ietf-gnap-resource-servers.md
jricher: Yes, all these sections need to be more clearly defined, the goal of this PR is just to get the right separation in place, so that work like this point can continue in the right place.

adeinega: I am guessing that `resource_server` could be also provided by value. Is this right?

yaronf:
Folks, I have a strong preference for the new document to live in its own repo (see https://www.rfc-editor.org/rfc/rfc8874.html#name-repositories for pros and cons). Any objections?

draft-ietf-gnap-resource-servers.md
jricher: @adeinega Yes, that's the intent of the "authenticating the RS" section, it would be parallel to how clients are identified in the core document.

jricher:
@yaronf I'm fine with this being in its own repository or in the same one -- I've used both in different groups and each has its benefits and drawbacks. If the chairs set up the repo I can import the support code and check in this document in its own PR.

yaronf:
https://github.com/ietf-wg-gnap/gnap-resource-servers

jricher:
RS document now moved to https://github.com/ietf-wg-gnap/gnap-resource-servers

Issue #196, opened 2021-02-18
Proving Possession of a Key for the AS and RS


adeinega:
The _Proving Possession of a Key with a Request_ section explicitly states that

> Any keys presented by the client instance to the AS or RS MUST be validated as part of the request in which they are presented

1. Unfortunately, I don't see how a client can make use of the jws param (Attached JWS payload) to make calls to "regular" RSs. Normally, they won't be ready to accept the "application/jose" content type and the corresponding payload for the endpoint that accepts say "application/json" and "application/xml".

2. I also believe that this specification shouldn't require the payload (HTTP Body) to be included in the calculation of the JWS signature for RSs in case of Detached JWS payload. The reason is that the client (or any logic that validates access tokens) doesn't always have access to the payload and the bright example for this would be NGinx and [its auth_request module](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html). Note, this approach isn't limited in any way to NGinx and is extremely popular among various cloud providers. There are also other minor reasons say just huge payloads (like a movie upload) or even a real-time stream (WebRTC) and so on and so forth.

Both these items are totally fine and definitely should be required for any calls from a client to an AS though.

adeinega:
https://tools.ietf.org/html/draft-ietf-oauth-dpop-02 says this about the payload and HTTP headers

> DPoP does not directly ensure message integrity but relies on the TLS layer for that purpose.

and

> DPoP does not ensure the integrity of the payload or headers of requests.  The DPoP proof only contains claims for the HTTP URI and method, but not, for example, the message body or general request headers.
> 
> This is an intentional design decision intended to keep DPoP simple to use, but as described, makes DPoP potentially susceptible to replay attacks where an attacker is able to modify message contents and headers.  In many setups, the message integrity and confidentiality provided by TLS is sufficient to provide a good level of protection.
> 
> Implementers that have stronger requirements on the integrity of messages are encouraged to either use TLS-based mechanisms or signed requests.

jricher:
[...]

PR #250, opened 2021-04-15
Cryptographic Key Proofing Updates

### LINKED ISSUE #201: Include body hash in Detached JWS

### LINKED ISSUE #216: Simplify access token hash
There have been several questions about the `at_hash` parameter currently used to convey the hash of an access token in requests to the RS. Right now we use the definition from OpenID Connect:

https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken

In speaking with the DPoP editors I've proposed a simplified claim there:

https://github.com/danielfett/draft-dpop/pull/62/

It's functionally similar but doesn't require the array truncation step, and it also fixes to a single hash algorithm (SHA-256) instead of having to figure out which hash to use based on the external signing envelope. I think this change makes it both simpler to implement and to understand.

### LINKED ISSUE #218: Update HTTP Message Signature Formats
The HTTP Message Signatures draft has changed formats, our examples need to be updated to incorporate these changes.

### LINKED ISSUE #230: Update "htu" claim name in JOSE methods
DPoP defines `htu` as not containing path or query components, GNAP's JOSE methods use the full URL. The JOSE methods should move to a different claim name to avoid confusion with DPoP.

### LINKED ISSUE #231: Create verifier text
Key proof sections do not currently detail how to verify each method, including requirements for verifying fields, this needs to be added.

jricher:
Updates a number of syntacitcal and semantic requirements for key proofing mechanisms and makes the language more generic around the use of the mechanisms in the protocol and its extensions. 

- change "htu" claim to "uri" for JOSE-based methods, closes #230
- align token hash with latest dpop 'ath' claim, closes #216
- use body hash for detached JWS with body, closes #201
- update HTTP message signature syntax, closes #218
- include verification language, closes #231

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3474,6 +3474,45 @@ to either the access token's own key or, in the case of bearer tokens, the clien
 
 \[\[ [See issue #105](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/105) \]\]
 
+In the following sections, unless otherwise noted, the `RS256` JOSE Signature Algorithm is applied
+using the following RSA key (presented here in JWK format):
+
+~~~
+{
+    "kid": "gnap-rsa",
+    "p": "xS4-YbQ0SgrsmcA7xDzZKuVNxJe3pCYwdAe6efSy4hdDgF9-vhC5gjaRk\
+        i1wWuERSMW4Tv44l5HNrL-Bbj_nCJxr_HAOaesDiPn2PnywwEfg3Nv95Nn-\
+        eilhqXRaW-tJKEMjDHu_fmJBeemHNZI412gBnXdGzDVo22dvYoxd6GM",
+    "kty": "RSA",
+    "q": "rVdcT_uy-CD0GKVLGpEGRR7k4JO6Tktc8MEHkC6NIFXihk_6vAIOCzCD6\
+        LMovMinOYttpRndKoGTNdJfWlDFDScAs8C5n2y1STCQPRximBY-bw39-aZq\
+        JXMxOLyPjzuVgiTOCBIvLD6-8-mvFjXZk_eefD0at6mQ5qV3U1jZt88",
+    "d": "FHlhdTF0ozTliDxMBffT6aJVKZKmbbFJOVNten9c3lXKB3ux3NAb_D2dB\
+        7inp9EV23oWrDspFtvCvD9dZrXgRKMHofkEpo_SSvBZfgtH-OTkbY_TqtPF\
+        FLPKAw0JX5cFPnn4Q2xE4n-dQ7tpRCKl59vZLHBrHShr90zqzFp0AKXU5fj\
+        b1gC9LPwsFA2Fd7KXmI1drQQEVq9R-o18Pnn4BGQNQNjO_VkcJTiBmEIVT_\
+        KJRPdpVJAmbgnYWafL_hAfeb_dK8p85yurEVF8nCK5oO3EPrqB7IL4UqaEn\
+        5Sl3u0j8x5or-xrrAoNz-gdOv7ONfZY6NFoa-3f8q9wBAHUuQ",
+    "e": "AQAB",
+    "qi": "ogpNEkDKg22Rj9cDV_-PJBZaXMk66Fp557RT1tafIuqJRHEufSOYnsto\
+        bWPJ0gHxv1gVJw3gm-zYvV-wTMNgr2wVsBSezSJjPSjxWZtmT2z68W1DuvK\
+        kZy15vz7Jd85hmDlriGcXNCoFEUsGLWkpHH9RwPIzguUHWmTt8y0oXyI",
+    "dp": "dvCKGI2G7RLh3WyjoJ_Dr6hZ3LhXweB3YcY3qdD9BnxZ71mrLiMQg4c_\
+        EBnwqCETN_5sStn2cRc2JXnvLP3G8t7IFKHTT_i_TSTacJ7uT04MSa053Y3\
+        RfwbvLjRNPR0UKAE3ZxROUoIaVNuU_6-QMf8-2ilUv2GIOrCN87gP_Vk",
+    "alg": "RS256",
+    "dq": "iMZmELaKgT9_W_MRT-UfDWtTLeFjIGRW8aFeVmZk9R7Pnyt8rNzyN-IQ\
+        M40ql8u8J6vc2GmQGfokLlPQ6XLSCY68_xkTXrhoU1f-eDntkhP7L6XawSK\
+        Onv5F2H7wyBQ75HUmHTg8AK2B_vRlMyFKjXbVlzKf4kvqChSGEz4IjQ",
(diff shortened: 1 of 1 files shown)


jricher:
At this time, cryptographic and encoded content of examples has not been regenerated and recalculated.

jricher:
Closes #196

draft-ietf-gnap-core-protocol.md
```diff
@@ -3363,39 +3371,44 @@ to either the access token's own key or, in the case of bearer tokens, the clien
...
 htm (string)
-: The HTTP Method used to make this request, as an uppercase ASCII string.
+: The HTTP Method used to make this request, as an uppercase ASCII string. REQUIRED
 
-htu (string)
-: The HTTP URI used for this request, including all path and query components.
+uri (string)
+: The HTTP URI used for this request, including all path and query components and no fragment component. REQUIRED
 
 ts (integer)
```
fimbault, line 3394: cat? created at

draft-ietf-gnap-core-protocol.md
```diff
@@ -3696,12 +3707,22 @@ fHI6kqm3NCyCCTihe2ck5RmCc5l2KBO/vAHF0ihhFOOOby1v6qbPHQcxAU6rEb907
 
 ~~~
 
+The verifier compares the TLS client certificate presented during 
+mutual TLS negotiation to the expected key of the signer. Since the
+TLS connection covers the entire message, there are no additional
+requirements to check.
+
+Note that in many instances, the verifier will not do a full certificate
```
fimbault, line 3715: `some instances` would be more appropriate?

draft-ietf-gnap-core-protocol.md
```diff
@@ -3778,28 +3792,56 @@ B7_8Wbw4ttzbMS_doJvuDagW8A1Ip3fXFAHtRAcKw7rdI4_Xln66hJxFekpdfWdiPQddQ6Y
...
+When the request is bound to an access token, the covered content
+MUST also include:
+
+authorization:
+: The Authorization header used to present the access token as discussed in
+{{use-access-token}}.
+
+Other covered content MAY also be included.
+
+If the singer's key presented is a JWK, the `keyid` parameter of the signature MUST be set
```
fimbault, line 3826: singer -> signer ;-)

fimbault (commented):
Looks fine, but tbh I must test to be sure

draft-ietf-gnap-core-protocol.md
jricher: No you see this method is only for musical APIs ;-)

draft-ietf-gnap-core-protocol.md
jricher: If we were going to change this I'd rather just use `created` to be honest. The three-letter constraints of JWT don't have a lot of bearing here.

draft-ietf-gnap-core-protocol.md
jricher: I think "many" is the right level here, to be honest -- since the protocol has lots of support for key-introduction. Either way it's just an implementation note, and we'll need security considerations on trusted/untrusted certificates and keys.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3474,27 +3485,30 @@ typ (string)
...
 
-at_hash (string)
-: When a request is bound to an access token, the access token hash value. Its value is the 
-    base64url encoding of the left-most half of the hash of the octets of the ASCII representation of the 
-    `access_token` value, where the hash algorithm used is the hash algorithm used in the `alg` 
-    header parameter of the JWS's JOSE Header. For instance, if the `alg` is `RS256`, hash the `access_token` 
-    value with SHA-256, then take the left-most 128 bits and base64url encode them. 
+ath (string)
+: When a request is bound to an access token, the access token hash value. The value MUST be the
+  result of a Base64url encoding (with no padding) the SHA-256 digest
```
yaronf, line 3496: *of* the SHA-256

draft-ietf-gnap-core-protocol.md
yaronf: I don't like the way this paragraph is worded. We are giving implementors a license to use self-signed client certs, without giving them guidance on when this is appropriate and when it is not. IMO by default the verifier MUST do full cert chain validation, and there are some cases when it MAY skip it.

jricher: I disagree with that precedence and with the normative statements -- since GNAP allows for key introduction both at runtime and during a pre-registration, the chain validation only fits limited deployment use cases where the AS has a CA for managing client certs. Also we're talking about validating the _client_ cert here, not the server cert. We can probably benefit from the language in RFC8705 (OAuth MTLS) which has similar considerations. 

It might make more sense to have this discussion in the security considerations as opposed to an aside here, though.

yaronf: To the best of my knowledge, MTLS only (mostly?) makes sense if you have a CA somewhere, otherwise it is not very scalable. And if any normative language is required, the Security Considerations is not the right place for it.

jricher: Traditional MTLS only scales with a CA because of its dependence on PKI. The trust root is different with this protocol because it works at a different layer.

Just about all of the deployments of OAuth MTLS I've personally been involved in haven't used PKI or CA's. Instead, client apps create and register a cert. This cert gets compared to the registered cert on the connection at the application layer, not at the TLS layer. On the TLS layer it's just assuring that the certificates in use in the connections are the ones indicated. It's a separate question whether or not those certificates are trusted or represent a specific known piece of software, and that question isn't answered at the TLS layer.

It's really analogous to having a non-PKI key and using a different signing mechanism. We want to have both options here for MTLS, and it makes sense to have guidance on the different modalities.

yaronf: OK, makes sense. I was thinking of service mesh deployments (e.g. Istio) where you do have a small local CA.

Please create an issue for moving this discussion to Sec Cons then.

jricher: A mini-CA is definitely something the AS could use to manage and limit the keys/certs it sees from clients, especially since cert-based keys could be used for the HTTP Signing or JOSE methods, too. 

New issue created to track this at #253

draft-ietf-gnap-core-protocol.md
jricher: changed to `created`

jricher:
Cryptographic examples have all been updated.

Issue #244, opened 2021-04-08
Refactoring the internals of access request


fimbault:
A few questions/suggestions around section 8 (Resource Access Rights): 
- type: considering the constraints (no collision), I believe the term namespace would be better suited. Since we explain a URI is prefered in real life scenarii, we might update the examples accordingly
- locations singular or plural: what's the rationale for an array of locations? Do we see that as a common scenario (with exactly the same parameters)? Wouldn't a single value fit most cases, as it's always possible to use multiple access tokens (2.1.2) anyways?
- actions: seems ok. Probably we should update the definition of right to `ability given to a subject to perform a given action on a resource under the control of an RS` (instead of operation)
- datatypes: seems to me that we should refer to an array of resources (possibly we could then have types and objects inside, but basic strings seem more generic and simpler). Depending on the use case, objects could include identifiers, absolute or relative paths, wildcards and specific semantics. 
- we're currently missing the possibility to request attributes (cf #233), which would be an array of strings to refer to groups, roles, etc. (again here, strings are enough I think, an implementer could always distinguish using an appropriate scheme, ex r_admin and g_club in the following example if he really wants to). Attributes would be optional.

(I'm not considering here the domain specific extensions, like geolocation, currency, etc. / btw, shouldn't this be in a dedicated subsection and not in every example?)

Example of proposed json for the access request:

```
{
    "access_token": [
    {
        "label": "fun_1234",
        "access": 
        [
            {
                "namespace": "https://as1.example.dev/media-api/v2/",               
                "location": "https://mediaserver.example.com",  
                "actions": [ "list", "view", "delete" ],
                "attributes": [ "admin", "club" ],
                "resources": [ "photo/*", "video/clip.mp4" ]
            },
            "dolphin"        
        ]
    }, 
    {
        "label": "profit_6789",
        "access": 
        [
            {
                "namespace": "https://as1.example.dev/bank-api/v1/",               
                "location": "https://bankserver.example.com",  
                "actions": [ "check_statement", "withdraw" ],  
                "attributes": [ "accountant", "euro" ],
                "resources": [ "account-42" ]
            },
            "expense"        
        ]

    }
]}

```

[...]

PR #259, opened 2021-05-11
add privileges field

jricher:
Addresses #244


yaronf:
I would love to see an example, i.e. a full access request that demonstrates whether/how this interacts with other properties of the request.

Also note that this is subtly different from the use of the word "privileges" in the Terminology (twice).

jricher:
@yaronf the editors discussed the need for adding an example that includes this field and decided we'd target a separate PR where we could clean up the examples more comprehensively instead of just pasting it in. 

The terminology alignment is noted, @fimbault do you think the definitions need to be tweaked a little?

fimbault:
Yes. I'll review the terminology.

PR #261, opened 2021-05-14
re-import rs-first discovery method from gnap-rs, closes #260

### LINKED ISSUE #118: RS www-authenticate response
§10.4 Requesting a Resource With Insufficient Access: Editor's note: 

This borrows heavily from UMA 2's "distributed authorization" model and, like UMA, might be better suited to an extension than the core protocol.

### LINKED ISSUE #260: Move client-facing sections back from RS draft
client requests to RS w/o access token should be covered in core (client-facing protocol)

See #118 also

jricher:
Closes #260 
Closes #118

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4765,9 +4765,6 @@ Detached-JWS: ejy0...
 If issued, the resulting access token would contain sufficient access to be used
 at both referenced resources.
 
-\[\[ [See issue #118](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/118) \]\]
-
-
 # Acknowledgements {#Acknowledgements}
 
 The editors would like to thank the feedback of the following individuals for their reviews,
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -4703,6 +4712,59 @@ mechanism, but the AS also returns other proofing methods, then the AS
...
+        "access": [
+            "FWWIKYBQ6U56NL1",
+            "dolphin-metadata"
+        ]
+    },
+    "client": "KHRS6X63AJ7C7C4AZ9AO"
+}
+~~~
+
+If issued, the resulting access token would contain sufficient access to be used
```
fimbault, line 4765: Seems ok, but we might be clearer of the "if issued" condition (again that's a decision made by the AS, which can use whichever rules make sense)

fimbault (commented):
Overall seems fine

draft-ietf-gnap-core-protocol.md
jricher: This is echoing language that's used elsewhere in the spec, which was suggested during a text review. If it needs to be changed, we should change it consistently. I personally think it's fine as-is since it keeps things simple in the context -- we don't care how or why it was issued, just whether it was.

fimbault: Yes this can be checked globally at a later stage, just to make sure everything is consistent

Issue #217, opened 2021-03-17
Add Roles Diagram


jricher:
Add an abstract diagram showing the connection between the different roles/parties in the protocol, without necessarily showing the temporal steps of the interactions between them.

PR #267, opened 2021-06-08
add role diagram

aaronpk:
adds a diagram illustrating the relationship between the roles in GNAP

for #217


draft-ietf-gnap-core-protocol.md
```diff
@@ -130,6 +130,41 @@ The parties in GNAP perform actions under different roles.
...
+│   Server    │            │   Server   │
+│             │◄─┐   ┌────►│            │
+└─────────────┘  │   │     └────────────┘
+       ▲         │   │
+       +         │   │
+       +         │   │
+       +         │   │
+       +         ▼   ▼
+       +       ┌──────────┐
+       +       │  Client  │
```
jricher, line 147: Do we want this to be a "client instance" instead?

jricher (changes requested):
Diagram itself is OK I think, but the line/arrow style is different from the other diagrams in the doc (this one seems to use higher ASCII/ANSI line characters instead of low-ASCII?). I'm fine with either but I think it should be consistent.

aaronpk:
Replaced with ascii chars!

jricher (approved):

yaronf:
We have been using the "goat" tool in one of my drafts to convert ASCII art to SVG. The results are very nice. You may want to consider this option.

* Tool: https://github.com/blampe/goat
* Makefile: https://github.com/yaronf/I-D/blob/main/STAR-Delegation/Makefile
* Source ASCII art: https://github.com/yaronf/I-D/tree/main/STAR-Delegation/art
* Plaintext rendering: https://www.ietf.org/archive/id/draft-ietf-acme-star-delegation-09.txt (see e.g. Fig. 11)
* HTML+SVG rendering: https://www.ietf.org/archive/id/draft-ietf-acme-star-delegation-09.html

PR #268, opened 2021-06-09
add grant endpoint URL to interaction hash to close mix-up proxy vulnerability

jricher:


draft-ietf-gnap-core-protocol.md
```diff
@@ -2678,19 +2678,22 @@ several kinds of session fixation and injection attacks. The AS MUST
...
+creates a hash string by concatenating the following values in the following order
+using a single newline (`\\n`) character to separate them:
+
+* the "nonce" value sent by the client instance in the [interaction "finish" section of the initial request](#request-interact-finish)
+* the AS's nonce value from [the interaction finish response](#response-interact-finish)
+* the "interact_ref" returned from the AS as part of the [interaction finish method](#interaction-finish)
+* the grant endpoint URL the client instance used to make its [initial request](#request)
+
 There is no padding or whitespace before or after any of the lines,
 and no trailing newline character.
```
yaronf, line 2690: The example below is misleading, because you can in fact see a trailing newline!

Maybe:
`VJLO6A4CAYLBXHTR0KRO\nMBDOFXG4Y5CVJCX821LH\n4IFWWIKYBC2PQ6U56NL1\nhttps://server.example.com/tx`

Or else clarify this point in the text.

yaronf:
Please ask the guy who submitted the vuln if there's any publication we can reference, maybe even the master's thesis as a WIP. They've earned it!

draft-ietf-gnap-core-protocol.md
jricher: The problem with formatting as a long line example using `\n` is that it would get wrapped with the linewrap functions, so we'd have trailing slashes in the middle and it would look weird. I think there's a way to add a caption to the codeblock, so that might be the best place to put it.

jricher:
@yaronf I've reached out to them if there's a good URL to reference, and added them to the ack section (I should have done that before!).

PR #269, opened 2021-06-25
split interaction methods discovery value

jricher:
The discovery fields for the interaction methods were still in a single list, even though those have now been split into start and finish sections.


fimbault (approved):
Yes, makes sense

aaronpk (approved):

Issue #146, opened 2020-12-16
Persistent Grant Identifier


fimbault:
Several contributors (Dave, Warren) asked whether there should be a **persistent identifier** of the grant.
See discussion on the mailing list as a starting point https://mailarchive.ietf.org/arch/msg/txauth/RB4u0uHS5SXSd3ngubBSuOkTC3s/ (Warren)
https://mailarchive.ietf.org/arch/msg/txauth/Txc1Y6HMSCA2-0sWP7h6GL3HbO4/ (Dave)

Before we dive into a potential technical solution, we should review the related use cases. Is it a problem if we don't have an identifier (or more precisely does it need to be exposed) ? What kind of operations would we do with it ?

fimbault:
I guess that would mean that we'd provide a uri specific to the grantid, for instance 
generated through https://github.com/ai/nanoid => grantID = "FF0faxylE7ece_WpZNmeW"

"continue": {
        "access_token": { xxxx }
        "uri": "https://server.example.com/FF0faxylE7ece_WpZNmeW"
    }

??

jricher:
The URI for the grant API needn't be the same as the identifier for the ongoing grant itself, and I don't see a good argument for forcing these two concepts to be tied together. Currently the access token is being used as an identifier, and I don't see a good reason for that being tied together either. Perhaps we need something as simple as a separate identifier?

fimbault:
Yes, it's a possibility, the question is: what will this identifier be used for, precisely? (see issue #62 too)

jricher:
So far that use case (extending and referencing an existing grant) is the only use case I've seen for an externalized identifier. That function was put in as an effort to support some of the advanced use cases that have been brought up. 

It's possible that this is a separable function that could go into an extension.

davidgtonge:
So use cases for grant identifiers, getting an existing grant status, etc. are:

1. Allowing a client to update an existing grant
2. Getting a status of a grant from the AS - e.g. how long left before access expires (this can be dynamic in some environments and is useful for the client to know)

Some work underway in this area (with a customer requirement): https://bitbucket.org/openid/fapi/src/master/Financial_API_Grant_Management.md
https://openbankinguk.github.io/read-write-api-site3/v3.1.7/resources-and-data-models/aisp/account-access-consents.html#get-account-access-consentsconsentid

fimbault:
Interesting. The grant management API in the fapi document doesn't allow updates, only queries and revoke, though.

[...]

Issue #131, opened 2020-11-24
Generic claim extension mechanism


fimbault:
Related to #64 (OpenID connect) but able to support also other schemes (e.g. VC), and related query syntax
Most probably by extensions

fimbault:
To revisit when we have something concrete

jricher:
This discussion gets us close into the identity protocol space, and the method for defining this should be a part of the extensions discussion in #134

fimbault:
This could help https://www.slideshare.net/TorstenLodderstedt/openid-for-verifiable-credentials-iiw-35 

Seems fairly straightforward to extend assertions formats and responses, although I'm wondering whether we should also include a type for assertions

PR #270, opened 2021-06-25
Trim features.

jricher:
This removes the "capabilities" and "existing_grant" portions of the request, as well as related aspects from discovery and responses. 

This also removes the separate "instance_id" field for the "client" part of the request, instead relying only on collapsing to a single value instead of an otherwise empty object.

These could still be added back in with extensions at some point in the future.
﻿

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1227,24 +1227,11 @@ within the `client` request, and other mechanisms.
 ### Identifying the Client Instance by Reference {#request-instance}
 
 If the client instance has an instance identifier that the AS can use to determine
-appropriate key information, the client instance can send this value in the `instance_id`
-field. The instance identifier MAY be assigned to a client instance at runtime
+appropriate key information, the client instance can send this instance 
+identifier as a direct reference value in lieu of the `client` object.
+The instance identifier MAY be assigned to a client instance at runtime
 through the {{response-dynamic-handles}} or MAY be obtained in another fashion,
 such as a static registration process at the AS.
-
-instance_id (string)
-: An identifier string that the AS can use to identify the
-    particular instance of this client software. The content and structure of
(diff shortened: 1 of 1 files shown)


fimbault (commented):
Personally in favor of that change, although i think it would make sense to make 2 separate PRs for that.
Regarding existing_grant, do we expect some impact on #146?

jricher:
Yes, "existing_grant" was the only driving reason behind having a persistent grant identifier. It's not needed for any of the other maintenance activities since you're given a URL and access token directly. My personal thought is that the identifier and existing_grant can be written together as an extension -- if you get the identifier you can use the existing grant, but you can't without it. That way it doesn't burden the rest of the system.

yaronf:
I would have preferred a separate PR for "existing grant" vs. "capabilities".

Specifically, I suggest to reconsider "capabilities".

First, there will be extensions in the future (especially the more features we trim from the base protocol), and it's good to have an extension mechanism from the get go.

Second, in security protocols it's useful to mark certain extensions as _critical_, and this can only be enforced if we have a "capabilities" field in the base protocol.

jricher:
For `capabilities`, I believe we would have a better idea of what the extension mechanism would be used for if we had any extensions that used it. I still like the feature, personally, but it doesn't make sense to have an extension point in place with nothing to hang from it -- it begs the question of whether it's the right extension point to begin with.

The rest of the protocol already has huge amounts of extensibility built in, in ways that don't need this feature. I think the group should revisit the topic but without this piece getting in the way of how we think about it when we do.

yaronf:
> I think the group should revisit the topic but without this piece getting in the way of how we think about it when we do.
> 
> We can plan to discuss all of the removed features at the next meeting.

So, should we have an issue to keep track of extensibility? And specifically with the requirement to enable future critical extensions?

jricher:
@yaronf #134 and #131 both cover different aspects of this, in my mind.

PR #271, opened 2021-07-05
trim dpop

fimbault:
remove OAuth DPoP binding. Described why on IETF mailing list: 
only works for asymmetric keys, requires key be presented in the header (duplicating information from GNAP messages). It was never meant to be a general purpose signing mechanism, though the FAPI group in OIDF is considering it as an option in current proposed work.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1,5 +1,4 @@
 ---
-
 title: 'Grant Negotiation and Authorization Protocol'
 docname: draft-ietf-gnap-core-protocol-latest
 category: std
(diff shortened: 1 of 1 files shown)


jricher (changes requested):
This PR has a lot of spurious formatting changes, including a number of things that break the YAML header at the start of the file. It looks like your editor might have auto-formatted the text? Please re-submit with only the intended changes.

fimbault:
Ok that's weird (just removed some lines...), will have a look

fimbault:
Ok that was due to my rust analyser in vscode, fixed it now

jricher (approved):

Issue #113, opened 2020-11-13
Applicability of OAuth PoP


jricher:
§8.6 OAuth Proof of Possession (PoP): Editor's note: 

This is a stale draft from the OAuth working group, but it does at least provide some basic functionality for protecting HTTP messages with a signature. This work is likely to be subsumed by the general-purpose HTTP message signature mechanism in Section 8.5.

jricher:
closed by #272

Issue #112, opened 2020-11-13
Use of OAuth PoP without access tokens


jricher:
§8.6 OAuth Proof of Possession (PoP): Editor's note: 

This is in contradiction to the referenced spec which makes this field mandatory.

jricher:
closed by #272

PR #272, opened 2021-07-05
trim oauthpop

fimbault:
remove OAuth DPoP binding. Described why on IETF mailing list:
OAuth PoP: expired draft, due to be replaced with new draft based on HTTP Message Signatures.

Also closing #112 and #113


jricher (changes requested):
Same issues as with #271, please resubmit with only intended changes.

fimbault:
Removed autoindent in my editor to fix the issue

jricher (approved):

PR #273, opened 2021-07-07
refactor token response details to use flags array

aaronpk:
...to match the flags array in the token request.

While this is technically a breaking change, I think we should consider this editorial as this should have been changed when we made the original change to use a flags array in the request in #162.


draft-ietf-gnap-core-protocol.md
```diff
@@ -1955,24 +1946,42 @@ key (object / string)
...
     to sign the request.
 
-durable (boolean)
+flags (array of strings)
+: OPTIONAL. A set of flags that represent attributes or behaviors of the access token
+    issued by the AS.
+
+The values of the `flags` field defined by this specification are as follows:
+
+bearer
```
jricher, line 1955:
```suggestion
"bearer"
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1955,24 +1946,42 @@ key (object / string)
...
+: This flag indicates whether the token is bound to the client instance's key.
+    If the `bearer` flag is present, the access token is a bearer token, and the `key`
+    field in this response MUST be omitted. If the `bearer` flag is omitted and the `key` field
+    in this response is omitted, the token is bound the [key used by the client instance](#request-client) 
+    in its request for access. If the `bearer` flag is omitted, and the `key` field is present,
+    the token is bound to the key and proofing mechanism indicated in the `key` field.
+
+durable
```
jricher, line 1963:
```suggestion
"durable"
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1955,24 +1946,42 @@ key (object / string)
...
-    in the issuance of new access tokens. If this flag is set to the boolean value `false`
-    or is omitted, the client instance can anticipate a given access token
+    in the issuance of new access tokens. If this flag is omitted, the client 
+    instance can anticipate a given access token
     will stop working after token rotation or grant request modification.
     Note that a token flagged as `durable` can still expire or be revoked through
     any normal means.
 
-split (boolean)
+split
```
jricher, line 1974:
```suggestion
"split"
```

jricher (changes requested):
Looks good - the two "bound" examples in the general response section need to be updated with the flag syntax. Nit: I'd recommend putting strings around the values for the flag definitions in the token request/response like we do in the interaction section. I added suggestions for the response but couldn't make suggestions for the request since it wasn't edited here.

aaronpk:
> Nit: I'd recommend putting strings around the values for the flag definitions in the token request/response like we do in the interaction section. I added suggestions for the response but couldn't make suggestions for the request since it wasn't edited here.

You know, I had that thought, and I checked the request section and it doesn't have quotes. I'm going to add the quotes to both parts.

> Looks good - the two "bound" examples in the general response section need to be updated with the flag syntax.

This ended up with some other language changing, not just the examples, so let me know what you think.

jricher (approved):

PR #294, opened 2021-07-19
add editorconfig

aaronpk:
this PR adds an `.editorconfig` file and also cleans up the markdown file to remove trailing spaces from lines.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -37,9 +37,9 @@ normative:
        title: Recommendations for Secure Use of Transport Layer Security (TLS) and Datagram Transport Layer Security (DTLS)
        date: May 2015
        author:
-         - 
+         -
            ins: Y. Sheffer
-         - 
+         -
            ins: R. Holz
          -
            ins: P. Saint-Andre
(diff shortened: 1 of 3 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -173,27 +173,27 @@ Legend
...
 Authorization Server (AS)
-: server that grants delegated privileges to a particular instance of client software in the form of access tokens or other information (such as subject information). 
+: server that grants delegated privileges to a particular instance of client software in the form of access tokens or other information (such as subject information).
 
 Client
-: application operated by an end-user that consumes resources from one or several RSs, possibly requiring access privileges from one or several ASs. 
+: application operated by an end-user that consumes resources from one or several RSs, possibly requiring access privileges from one or several ASs.
 
     Example: a client can be a mobile application, a web application, etc.
-
```
jricher, line 182: We should be careful of lines like this -- whitespace-only lines have semantic meaning in certain context in Markdown (particularly in list and block continuation). So do two spaces at the end of a line (linebreak without paragraph break). Is there a way to cover these odd cases by the editorconfig, which otherwise looks good?

PR #305, opened 2021-08-27
Use opaque identifier instead of user handle

### LINKED ISSUE #51: User reference as an assertion
§2.4.1 Identifying the User by Reference: Editor's note: 

We might be able to fold this function into an unstructured user assertion reference issued by the AS to the RC. We could put it in as an assertion type of "gnap_reference" or something like that. Downside: it's more verbose and potentially confusing to the client developer to have an assertion-like thing that's internal to the AS and not an assertion.

### LINKED ISSUE #197: Requesting User Information
In section 2.4 (Requesting User Information), the first sentence starts with:

     If the client instance is requesting information about the RO from the AS, ...

From the title, the acronym "RO" should be changed into "end-user". The case where the RO is also the end-user is a specific case. 

This leads to the following change:

     If the client instance is requesting information about the end-user from the AS, ...

In the following sentences of this section, RO should be changed into end-user.

### LINKED ISSUE #302: Use opaque identifer instead of user handle
With the adoption of the `opaque` subject identifier format, we no longer need to have a GNAP-specific `user_handle` value and can instead leverage this more standard field. Alternatively, we could have a GNAP-specific subject identifier format, instead of a separate `user_handle` field.

fimbault:
Opaque identifiers provide a generic reference, so we can simplify the text. 

Closes #302 (mainly), and as a byproduct also closes #197 and closes #51
- removed user_handle from main response in section 3
- changed title of section 3.4 (User -> Subject) and moved the text that requires RO = end-user at the top of the section, for clarity
- changed name (user_handle -> user) and wording in section 3.5. Maybe we could also simplify the title "Returning Dynamically-bound Reference Handles" into "Returning Reference Handles" (didn't do it yet).

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -77,7 +77,7 @@ normative:
         -
           ins: C. Mortimore
 
---- abstract
+​--- abstract
 
 GNAP defines a mechanism for delegating authorization to a
 piece of software, and conveying that delegation to the software. This
@@ -85,7 +85,7 @@ delegation can include access to a set of APIs as well as information
 passed directly to the software.
 
 
---- middle
+​--- middle
 
 # Introduction
 
@@ -167,7 +167,7 @@ on the role by the overall protocol.
 Legend
 
 + + + indicates interaction between a human and computer
------ indicates interaction between two pieces of software
+​----- indicates interaction between two pieces of software
 ~ ~ ~ indicates a potential equivalence or out-of-band communication between roles
 
 ~~~
@@ -333,7 +333,7 @@ an access token to call an RS.
 
 Legend
 + + + indicates a possible interaction with a human
------ indicates an interaction between protocol roles
+​----- indicates an interaction between protocol roles
 ~ ~ ~ indicates a potential equivalence or out-of-band
         communication between roles
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -2321,18 +2315,15 @@ instance_id (string)
...
             described in {{request-instance}}.
 
-user_handle (string)
-: A string value used to represent the current
-            user. The client instance can use in a future request, as described in
-            {{request-user-reference}}.
-
+user (string)
+: A string value used to represent the current user, that can be provided to a future request, as described in
+            {{request-user-reference}}. When the end-user is the RO, the handle is equivalent to the opaque subject identifier.
```
jricher, line 2320: I think we should drop this response field entirely and instead only use the `subject` response with an opaque identifier.

jricher (changes requested):
I don't think this goes far enough -- I am thinking that with the opaque identifiers we can now simply return a `subject` with a format of `opaque` that the client can be allowed to send in a follow-up request if it wants to. This way they all get treated equally, there's no special field anymore.

draft-ietf-gnap-core-protocol.md
fimbault: Thanks Justin. i was hesitant to get that far, but that would be fine for me.

PR #307, opened 2021-09-22
initial draft of privacy considerations section

aaronpk:


jricher (approved):

Denisthemalice:
Draft -07 states in section 13 (Privacy Considerations): 

> 13.  Privacy Considerations
> 
>    The privacy considerations in this section are modeled after the list
>    of privacy threats in [[RFC6973]], "Privacy Considerations for
>    Internet Protocols", and either explain how these threats are
>    mitigated or advise how the threats relate to GNAP.

Given this statement, it is unfortunate that the Guidelines (Section 7) given in RFC 6973 have not been followed 
and thus most questions have been left unanswered.

In particular, section 7.1 states:

> 7.1.  Data Minimization
> 
>   a.  Identifiers.  What identifiers does the protocol use for
>        distinguishing initiators of communications?  Does the protocol
>        use identifiers that allow different protocol interactions to be
>        correlated?  What identifiers could be omitted or be made less
>        identifying while still fulfilling the protocol’s goals?
> 
>    b.  Data.  What information does the protocol expose about
>        individuals, their devices, and/or their device usage (other than
>        the identifiers discussed in (a))?  To what extent is this
>        information linked to the identities of the individuals?  How
>        does the protocol combine personal data with the identifiers
>        discussed in (a)?
> 
>   c.  Observers.  Which information discussed in (a) and (b) is exposed
>        to each other protocol entity (i.e., recipients, intermediaries,
>        and enablers)?  Are there ways for protocol implementers to
>        choose to limit the information shared with each entity?  Are
>        there operational controls available to limit the information
>        shared with each entity?
> 
> e.  Persistence of identifiers.  What assumptions are made in the
>        protocol design about the lifetime of the identifiers discussed
>        in (a)?  Does the protocol allow implementers or users to delete
>        or replace identifiers?  How often does the specification
>        recommend deleting or replacing identifiers by default?  Can the
>        identifiers, along with other state information, be set to
>        automatically expire?
> 
>    f.  Correlation.  Does the protocol allow for correlation of
>        identifiers?  
> 

In particular, section 7.2 states:

> 7.2.  User Participation
> 
>    a.  User control.  What controls or consent mechanisms does the
>        protocol define or require before personal data or identifiers
>        are shared or exposed via the protocol?  If no such mechanisms or
>        controls are specified, is it expected that control and consent
>        will be handled outside of the protocol?
> 
>   c.  Control over sharing with intermediaries.  Does the protocol
>        provide ways for initiators to limit which information is shared
>        with intermediaries?  If not, are there mechanisms that exist
>        outside of the protocol to provide users with such control?  Is
>        it expected that users will have relationships that govern the
>        use of the information (contractual or otherwise) with those who
>        operate these intermediaries?
> 
>    d.  Preference expression.  Does the protocol provide ways for
>        initiators to express individuals’ preferences to recipients or
>        intermediaries with regard to the collection, use, or disclosure
>        of their personal data?
 

The current text is lacking to answer to the above questions.

**_Some comments on the proposed text_**

1/   The proposed text does not make any difference whether rights or attributes are supported by the RS to allow performing operations on a given resource under the control of an RS. However, major differences would need to be made.

They are detailed in draft-pinkas-gnap-core-protocol-00. See :

> 5.1.1.  Privacy Considerations for ABAC (Attribute Based Access Control) 
> 
> and
> 
> 5.1.2.  Privacy Considerations for CBAC (Capability Based Access Control)


2/   13.1.2.  Surveillance by the Authorization Server

> (...) When the authorization
>    server shares user information with the client instance, it needs to
>    make sure that it has the permission from that user to do so.

Unfortunately, the core protocol does not support any feature to achieve that goal. 

3/  The text continues with:
  
>  Additionally, as part of the authorization grant process, the
>    authorization server may be aware of which resource servers the
>    client intends to use an access token at.  However, it is possible to
>    design a system using GNAP in which this knowledge is not made
>    available to the authorization server, such as by avoiding the use of
>    the "locations" object in the authorization request.


Avoiding the use of the "locations" object is by no means a way to solve this issue. One year ago, a mechanism to hide to the AS to identity of the RS has been explained on the mailing list. It allows to target the access token to the intended RS while hiding to the AS the value of the identifier of the RS.

It is detailed in draft-pinkas-gnap-core-protocol-00. See the "hidden_url" field of the access token on page 9.

4/  The text continues with:

>    If the authorization server’s implementation of access tokens is such
>    that it requires a resource server call back to the authorization
>    server to validate them, then the authorization server will be aware
>    of which resource servers are actively in use and by which users and
>    which clients.  To avoid this possibility, the authorization server
>    would need to structure access tokens in such a way that they can be
>    validated by the resource server without notifying the authorization
>    server that the token is being validated.

This is insufficient: in addition, the AS SHALL not make available a token introspection end-point.

5/   Section 13.3 (Intrusion) deals with denial of service which has nothing to do with privacy.

6/   The last paragraph of this section states:

> The format and contents of subject identifiers are intentionally not defined by GNAP.

This leaves the door wide opened on how a RS can verify that an access token is associated (or not) 
with the right (short-term or long-term) user account.

This case is addressed in details in draft-pinkas-gnap-core-protocol-00. 
See the field "buid" which is a "binding user identifier" field of the access token on page 9.

7/ The text continues with:

> 13.4.1.  Correlation by Clients
> 
>    The biggest risk of correlation in GNAP is when an authorization
>    server returns stable consistent user identifiers to multiple different applications.
> 

Correlations by Clients is not a concern, but correlations by Resource Servers (section 13.4.2) is a major concern.

8/ The text continues with:

>    In some cases a resource server may not actually need to be able to
>    identify users, (such as a resource server providing access to a
>    company cafeteria menu which only needs to validate whether the user
>    is a current employee), so authorization servers should be thoughtful
>    of when user identifiers are actually necessary to communicate to
>    resource servers for the functioning of the system.

This is wrong. Since draft -07 considers that "The format and contents of subject identifiers are intentionally not defined by GNAP", 
it is unable to provide any solution.

This case is addressed in details in draft-pinkas-gnap-core-protocol-00. 
See the field "buid" which is a "binding user identifier" field of the access token on page 9.

PR #311, opened 2021-09-25
Fix label of arrow

pq2:
In step 2, the AS signals that interaction is needed and does not request access.


jricher:
Good catch, thank you!

PR #316, opened 2021-10-07
Replace "split field" with "split flag"

pq2:


jricher (approved):

jricher:
@pq2 The editors would like to give you credit for all these contributions, can you please give us your name as you'd like it to appear in the acknowledgements section? Thank you!

PR #340, opened 2021-11-01
Fix sorting of acknowledgements

pq2:


adeinega:
@jricher, interestingly enough that you as well as Fabian and Aaron are the editors of this specification. Thus, probably there is no need to be mentioned in this list, or the whole paragraph should be somehow rephrased.

PR #351, opened 2021-11-09
Add redirection status code security considerations

pq2:
Based on the [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.11)

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -5070,6 +5070,12 @@ redirect the RO to the client instance's redirect URI. Due to the use of status 
 user agent now transmits the RO's credentials to the client instance. A malicious client instance
 can then use the obtained credentials to impersonate the RO at the AS.
 
+Redirection away from the initial URL in an interaction session could also leak information found in that
+initial URL through the HTTP `Referer` header field, which would be sent by the user agent to the redirect
+target. To avoid such leakage, a server can first redirect to an internal interstitial page without any identifying
+or sensitive information on the URL before processing the request. When the user agent is ultimately 
+redirected from this page, no part of the original interaction URL will be found in the Referrer header.
+
 ## MTLS Message Integrity {#security-mtls}
 
 The [MTLS key proofing mechanism](#mtls) provides a means for a client instance to present a key 
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -5055,9 +5055,9 @@ As already described in {{I-D.ietf-oauth-security-topics}}, a server should neve
 status code to redirect a request that potentially contains user credentials. If an HTTP redirect
 is used for such a request, the HTTP status code 303 "See Other" should be used instead.
 
-The status code 307, as defined in the HTTP standard {{RFC7231}}, does not require the user agent
-to rewrite a POST request into a GET request, thus preserving the form data in the body of the POST
-request. In the HTTP standard {{RFC7231}}, only the status code 303 unambiguously enforces
+The status code 307, as defined in the HTTP standard {{RFC7231}}, requires the user agent
+to preserve the method and body of a request, thus submitting the body of the POST
+request to the redirect target. In the HTTP standard {{RFC7231}}, only the status code 303 unambiguously enforces
 rewriting the HTTP POST request to an HTTP GET request. For all other status codes, including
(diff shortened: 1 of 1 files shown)


fimbault:
Looks good, although there seems to be some whitespace issues

draft-ietf-gnap-core-protocol.md
```diff
@@ -5045,10 +5049,31 @@ software to look up the details of the pending request. Since this approach requ
...
 
+## Redirection Status Codes {#security-redirect-status-codes}
+
+As already described in {{I-D.ietf-oauth-security-topics}}, a server should never use the HTTP 307
+status code to redirect a request that potentially contains user credentials. If an HTTP redirect
+is used for such a request, the HTTP status code 303 "See Other" should be used instead.
+
+The status code 307, as defined in the HTTP standard {{RFC7231}}, does not require the user agent
+to rewrite a POST request into a GET request, thus preserving the form data in the body of the POST
+request. In the HTTP standard {{RFC7231}}, only the status code 303 unambiguously enforces
```
jricher, line 5060:
```suggestion
The status code 307, as defined in the HTTP standard {{RFC7231}}, requires the user agent
to preserve the method and body of a request, thus submitting the body of the POST
request to the redirect target. In the HTTP standard {{RFC7231}}, only the status code 303 unambiguously enforces
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5045,10 +5049,31 @@ software to look up the details of the pending request. Since this approach requ
...
+rewriting the HTTP POST request to an HTTP GET request. For all other status codes, including
```
jricher, line 5061:
```suggestion
rewriting the HTTP POST request to an HTTP GET request, which eliminates the POST body from the redirected request. For all other status codes, including
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5045,10 +5049,31 @@ software to look up the details of the pending request. Since this approach requ
...
+to resubmit the body.
+
+The use of status code 307 results in a vulnerability when using the
+[`redirect` interaction finish method](#response-interact-finish). With this method, the AS
+potentially prompts the RO to enter their credentials in a form that is then submitted back to the
+AS (using an HTTP POST request). The AS checks the credentials and, if successful, may directly
+redirect the RO to the client instance's redirect URI. Due to the use of status code 307, the RO's
+user agent now transmits the RO's credentials to the client instance. A malicious client instance
+can then use the obtained credentials to impersonate the RO at the AS.
+
```
jricher, line 5072:
```suggestion

Redirection away from the initial URL in an interaction session could also leak information found in that
initial URL through the HTTP Referrer header field, which would be sent by the user agent to the redirect
target. To avoid such leakage, a server can first redirect to an internal interstitial page without any identifying
or sensitive information on the URL before processing the request. When the user agent is ultimately 
redirected from this page, no part of the original interaction URL will be found in the Referrer header.

```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5358,7 +5383,7 @@ Throughout many parts of GNAP, the parties pass shared references between each o
     - Editorial updates.
 
 - -07
-    - Replace user handle by opaque identifier 
+    - Replace user handle by opaque identifier
```
jricher, line 5389:
```suggestion
        - Add security consideration on redirection processing
        - Replace user handle by opaque identifier
```

jricher (changes requested):
Please remove the changes to only whitespace so as not to bury the text contributions. Please see individual comments below for suggested new wording for the contributed text. Thank you!

pq2:
The changes to the whitespace are made automatically by any editor that respects the editorconfig file. I can undo them, but for anyone who wants to make changes in the future with an editor that respects editorconfig, those changes will then be made again automatically. Presumably these whitespace issues should be fixed by a separate commit in the main branch if the editorconfig file is to continue to be used.

draft-ietf-gnap-core-protocol.md
pq2: I'm not sure if it should rather be called "HTTP `Referer` header", since the official name of the header is a misspelling. But I'm also not sure what the conventions are for this. Couldn't the HTTP `Referrer-Policy` header also be used for this?

PR #357, opened 2021-12-06
Consistent use of "end-user" instead of "end user"

pq2:
Section 1.2 only defines the role of an `end-user`, so I thought this should be written consistently that way.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -414,7 +414,7 @@ Legend
     other methods. The RO approves the request for the client instance
     being operated by the end-user. Note that the RO and end-user are often
     the same entity in practice, and many of GNAP's interaction methods allow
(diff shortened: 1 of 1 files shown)


jricher (approved):
I think this is the correct move, but I'd like to hear what @aaronpk and @fimbault think about the changes.

fimbault:
Yes that's good

fimbault:
Upon review (thanks Aaron), it should be unhyphenated as a noun. We sure need to be consistent everywhere.
See https://en.wikipedia.org/wiki/End_user

pq2:
I have replaced the unambiguous uses of "end-user" as a noun with "end user". I am not sure about "end-user roles" in line 232 and "Exposure of end-user credentials" in line 4878. I left these lines as before (hyphenated). Someone else should look at these lines again.

PR #359, opened 2021-12-10
Update Unicode Handling

### LINKED ISSUE #352: Unicode In Document
Some unicode characters within the document break the build tools, specifically we see `Å` being changed to `Ae` in the rendered document.

jricher:
Closes #352﻿


draft-ietf-gnap-core-protocol.md
```diff
@@ -4588,7 +4588,7 @@ at both referenced resources.
 
 The editors would like to thank the feedback of the following individuals for their reviews,
 implementations, and contributions:
-Åke Axeland,
+{{{Åke Axeland}{Ake Axeland}}},
```
aaronpk, line 4591: What's going on here?

PR #360, opened 2021-12-10
Mitigate Cuckoo Token Attack

### LINKED ISSUE #354: Security Consideration: Cuckoo Token Mitigations
- Client instance uses different keys with each AS
    - Stolen token bound to different keys, RS will reject
- Client has strong binding between RS and AS used
    - Attacker can’t convince client to use “wrong” AS

jricher:
Added security considerations for cuckoo token, closes #354

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -5422,6 +5422,7 @@ Throughout many parts of GNAP, the parties pass shared references between each o
 
 - -09
     - Added security considerations on redirection status codes.
+    - Added security considerations on cuckoo token attack.
 
 - -08
     - Update definition for "Client" to account for the case of no end-user.
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -99,6 +99,13 @@ informative:
                 ins: Å. Axeland
             -
                 ins: O. Oueidat
+    HELMSCHMIDT2021:
+        target: 'tbd'
+        title: 'tbd'
+        date: 2021
+        author:
+            -
+                ins: F. Helmschmidt
 
 --- abstract
 
@@ -5302,6 +5309,36 @@ have a secure and correct XML parser. In addition to this, the SAML 2 specificat
 Signatures, which have their own implementation problems that need to be accounted for. Similar
 requirements exist for OpenID Connect's ID token, which is based on the JSON Web Token (JWT) format
 and the related JSON Object Signing And Encryption (JOSE) cryptography suite.
+
+## Stolen Token Replay {#security-cuckoo}
+
+If a client instance cam request tokens at multiple AS's, and the client instance uses the same keys
+to make its requests across those different AS's, then it is possible for an attacker to replay a
+stolen token issued by an honest AS from a compromised AS, thereby binding the stolen token to
+the client instance's key in a different context. The attacker can manipulate the client instance
+into using the stolen token at an RS, particularly at an RS that is expecting a token from the
+honest AS. Since the honest AS issued the token and the client instance presents the token with
+its expected bound key, the attack succeeds.
+
+This attack has several preconditions. In this attack, the attacker does not need access to the
+client instance's key and cannot use the stolen token directly at the RS, but the attacker is able
+to get the access token value in some fashion. The client instance also needs to be configured to
+talk to multiple AS's, including the attacker's controlled AS. Finally, the client instance needs
+to be able to be manipulated by the attacker to call the RS while using a token issued from the
+stolen AS. The RS does not need to be compromised or made to trust the attacker's AS.
+
+To protect agains this attack, the client instance can use a different key for each AS that it
+talks to. Since the replayed token will be bound to the key used at the honest AS, the
+uncompromised RS will reject the call since the client instance will be using the key used at
+the attacker's AS instead with the same token.
+
+Additionally, the client instance can keep a strong association between the RS and a specific AS
+that it trusts to issue tokens for that RS. This strong binding also helps against some forms of
+[AS mix-up attacks](#security-mixup). Managing this binding is outside the scope of GNAP core,
+but it can be managed either as a configuration element for the client instance or dynamically
+through [discovering the AS from the RS](#rs-request-without-token).
+
+The details of this attack are available in {{HELMSCHMIDT2021}} with additional discussion and considerations.
 
 # Privacy Considerations {#privacy}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -99,6 +99,13 @@ informative:
                 ins: Å. Axeland
             -
                 ins: O. Oueidat
+    HELMSCHMIDT2021:
```
pq2, line 102:
```suggestion
    HELMSCHMIDT2022:
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -99,6 +99,13 @@ informative:
...
+        target: 'tbd'
+        title: 'tbd'
+        date: 2021
```
pq2, line 105:
```suggestion
        date: 2022
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5303,6 +5313,36 @@ Signatures, which have their own implementation problems that need to be account
 requirements exist for OpenID Connect's ID token, which is based on the JSON Web Token (JWT) format
 and the related JSON Object Signing And Encryption (JOSE) cryptography suite.
 
+## Stolen Token Replay {#security-cuckoo}
+
+If a client instance cam request tokens at multiple AS's, and the client instance uses the same keys
```
pq2, line 5318:
```suggestion
If a client instance can request tokens at multiple ASs, and the client instance uses the same keys
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5303,6 +5313,36 @@ Signatures, which have their own implementation problems that need to be account
...
+stolen token issued by an honest AS from a compromised AS, thereby binding the stolen token to
+the client instance's key in a different context. The attacker can manipulate the client instance
+into using the stolen token at an RS, particularly at an RS that is expecting a token from the
+honest AS. Since the honest AS issued the token and the client instance presents the token with
+its expected bound key, the attack succeeds.
+
+This attack has several preconditions. In this attack, the attacker does not need access to the
+client instance's key and cannot use the stolen token directly at the RS, but the attacker is able
+to get the access token value in some fashion. The client instance also needs to be configured to
+talk to multiple AS's, including the attacker's controlled AS. Finally, the client instance needs
```
pq2, line 5329:
```suggestion
talk to multiple ASs, including the attacker's controlled AS. Finally, the client instance needs
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5303,6 +5313,36 @@ Signatures, which have their own implementation problems that need to be account
...
+to be able to be manipulated by the attacker to call the RS while using a token issued from the
+stolen AS. The RS does not need to be compromised or made to trust the attacker's AS.
+
+To protect agains this attack, the client instance can use a different key for each AS that it
```
pq2, line 5333:
```suggestion
To protect against this attack, the client instance can use a different key for each AS that it
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5303,6 +5313,36 @@ Signatures, which have their own implementation problems that need to be account
...
+uncompromised RS will reject the call since the client instance will be using the key used at
+the attacker's AS instead with the same token.
+
+Additionally, the client instance can keep a strong association between the RS and a specific AS
+that it trusts to issue tokens for that RS. This strong binding also helps against some forms of
+[AS mix-up attacks](#security-mixup). Managing this binding is outside the scope of GNAP core,
+but it can be managed either as a configuration element for the client instance or dynamically
+through [discovering the AS from the RS](#rs-request-without-token).
+
+The details of this attack are available in {{HELMSCHMIDT2021}} with additional discussion and considerations.
```
pq2, line 5344:
```suggestion
The details of this attack are available in {{HELMSCHMIDT2022}} with additional discussion and considerations.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1288,6 +1295,9 @@ given at runtime. Additional fields sent during a request but not present in a p
 client instance record at the AS SHOULD NOT be added to the client's pre-registered record.
 See additional considerations regarding client instance impersonation in {{security-impersonation}}.
 
+A client instance that is capable of talking to multiple AS's SHOULD use a different key for each
```
pq2, line 1298:
```suggestion
A client instance that is capable of talking to multiple ASs SHOULD use a different key for each
```
I'm not sure if the plural of `AS` should be `ASs` or `AS's`. `ASs` is used twice in the document, while `AS's` is also used twice.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5303,6 +5313,36 @@ Signatures, which have their own implementation problems that need to be account
...
+client instance's key and cannot use the stolen token directly at the RS, but the attacker is able
+to get the access token value in some fashion. The client instance also needs to be configured to
+talk to multiple AS's, including the attacker's controlled AS. Finally, the client instance needs
+to be able to be manipulated by the attacker to call the RS while using a token issued from the
+stolen AS. The RS does not need to be compromised or made to trust the attacker's AS.
+
+To protect agains this attack, the client instance can use a different key for each AS that it
+talks to. Since the replayed token will be bound to the key used at the honest AS, the
+uncompromised RS will reject the call since the client instance will be using the key used at
+the attacker's AS instead with the same token.
```
pq2, line 5336:
```suggestion
the attacker's AS instead with the same token.
When the MTLS key proofing method is used, a client instance can use self-signed
certificates to use a different key for each AS that it talks to, as discussed in
{{security-mtls-patterns}}.
```

pq2 (changes requested):

draft-ietf-gnap-core-protocol.md
jricher: I'd just put this in as a placeholder until you can publish it, we'll obviously add all the information when we can. :)

draft-ietf-gnap-core-protocol.md
jricher: I don't think there's a good way to do it, to be honest, but I'm not a fan of "ASs" because it looks like a typo of "ass". Still you're right that the doc should be consistent, so we'll want to fix that eventually -- but I'm leaving these all as-is for now.

draft-ietf-gnap-core-protocol.md
jricher:
```suggestion
If a client instance can request tokens at multiple AS's, and the client instance uses the same keys
```

PR #362, opened 2021-12-14
expand section on key references

### LINKED ISSUE #154: Common means to dereference key references
> I disagree: when used between the Client and AS this is clearly required for interop.

_Originally posted by @yaronf in https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/152#discussion_r548538220_

aaronpk:
closes #154

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3493,6 +3493,16 @@ Keys referenced in this manner MUST be bound to a single proofing mechanism.
 
 The means of dereferencing this value are out of scope for this specification.
 
+Commonly, key references are created by the AS and are not necessarily needed
+to be dereferencable by the client. These types of key references are an
+internal reference to the AS, such as an identifier of a record in a database.
+
+In other applications, it can be useful to use key references that are resolvable
+by both clients and ASs, which could be accomplished by e.g. a client publishing
+a public key at a URL. For interoperability, this method could later be described
+as an extension.
+
+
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3493,6 +3493,16 @@ Keys referenced in this manner MUST be bound to a single proofing mechanism.
...
+Commonly, key references are created by the AS and are not necessarily needed
+to be dereferencable by the client. These types of key references are an
+internal reference to the AS, such as an identifier of a record in a database.
+
+In other applications, it can be useful to use key references that are resolvable
+by both clients and ASs, which could be accomplished by e.g. a client publishing
+a public key at a URL. For interoperability, this method could later be described
+as an extension.
+
+
```
jricher, line 3505: I'd recommend that this all be the same paragraph, so that the lead-in doesn't get lost from the explanation. But I think that overall this is good!

```suggestion
The means of dereferencing this value are out of scope for this specification.
Commonly, key references are created by the AS and are not necessarily needed
to be dereferencable by the client. These types of key references are an
internal reference to the AS, such as an identifier of a record in a database.
In other applications, it can be useful to use key references that are resolvable
by both clients and ASs, which could be accomplished by e.g. a client publishing
a public key at a URL. For interoperability, this method could later be described
as an extension.


```

jricher (approved):
I think the text looks good, left one note on formatting/structure but it's not a showstopper.

PR #366, opened 2021-12-23
Token Rotation Improvements

### LINKED ISSUE #101: Invalidating previous access tokens during rotation

### LINKED ISSUE #102: Require inclusion of token management URI in rotation response
§6.1 Rotation the Access Token: Editor's note: 

The net result is that the client's always going to use the management URL that comes back. But should we let the server omit it from the response if it doesn't change? That seems like an odd optimization that doesn't help the client.

jricher:
- update token rotation response and requirements
- add considerations for the difficulties of rotating and revoking stateless tokens
- add considerations on rotation in unreliable networks

Closes #101
Closes #102

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3336,8 +3336,8 @@ honor a rotation request for an access token that has been revoked or otherwise 
 
 If the token is validated and the key is appropriate for the
 request, the AS MUST invalidate the current access token associated
-with this URL, if possible.
-\[\[ [See issue #101](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/101) \]\]
+with this URL, if possible. Note that stateless access tokens can make proactive
+revocation difficult within a system, see {{security-stateless-tokens}}.
 
 The AS responds with an HTTP 200 with a JSON body consisting of the rotated access token
 in the `access_token` field described in {{response-token-single}}. The value of the
@@ -3422,6 +3422,8 @@ means, the AS SHOULD honor the revocation request to
 the token management URL as valid, since the end result is still the token
 not being usable.
 
+Note that some deployment decisions like self-contained stateless access tokens can make
+propagating revocation difficult, see {{security-stateless-token}}.
 
 # Securing Requests from the Client Instance {#secure-requests}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3321,25 +3327,31 @@ The AS validates that the token presented is associated with the management
 URL, that the AS issued the token to the given client instance, and that
 the presented key is appropriate to the token.
 
-If the access token has expired, the AS SHOULD honor the rotation request to
-the token management URL since it is likely that the client instance is attempting to
+Note that in many cases, the access token will have expired for regular use. To facilitate
+token rotation, the AS SHOULD honor the rotation request of the expired access token
+since it is likely that the client instance is attempting to
 refresh the expired token. To support this, the AS MAY apply different lifetimes for
```
fimbault, line 3333: actually, what we're saying here is that the lifetime for token management might be longer that for its use

draft-ietf-gnap-core-protocol.md
```diff
@@ -3321,25 +3327,31 @@ The AS validates that the token presented is associated with the management
...
 the use of the token in management vs. its use at an RS. An AS MUST NOT
-honor a rotation request for an access token that has been revoked, either by
-the AS or by the client instance through the [token management URI](#revoke-access-token).
+honor a rotation request for an access token that has been revoked or otherwise disabled.
```
fimbault, line 3335: what do we mean by disabled ?

draft-ietf-gnap-core-protocol.md
```diff
@@ -3409,6 +3422,8 @@ means, the AS SHOULD honor the revocation request to
 the token management URL as valid, since the end result is still the token
 not being usable.
 
+Note that some deployment decisions like self-contained stateless access tokens can make
```
fimbault, line 3425: isn't it a bit redundant with what is already said above ? Note that stateless access tokens can make proactive
revocation difficult within a system, see {{security-stateless-tokens}}.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5354,6 +5369,61 @@ through [discovering the AS from the RS](#rs-request-without-token).
...
+should probably not offer token revocation functionality to client instances, since the client
+instance's calls to such an endpoint is effectively meaningless. However, a client instance calling
+the token revocation function will also throw out its copy of the token, so such a placebo endpoint
+might not be completely meaningless. Token rotation similarly difficult because the AS has to
+revoke the old access token after a rotation call has been made. If the access tokens are
+completely self-contained and non-revocable, this means that there will be a period of time during
+which both the old and new access tokens are valid and usable, which is an increased security risk
+for the environment.
+
+These problems can be mitigated by keeping the validity time windows of self-contained access tokens
```
fimbault, line 5394: probably we should say that the mitigations should be applied for specific use cases, that aren't too sensitive

fimbault (commented):
a few minor comments

draft-ietf-gnap-core-protocol.md
jricher: Generically, just "made unusable for any reason whatsoever". Revocation is the most concrete -- and permanent -- version of that.

draft-ietf-gnap-core-protocol.md
jricher: I believe you're correct, this paragraph can be dropped.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3409,6 +3422,8 @@ means, the AS SHOULD honor the revocation request to
 the token management URL as valid, since the end result is still the token
 not being usable.
 
+Note that some deployment decisions like self-contained stateless access tokens can make
+propagating revocation difficult, see {{security-stateless-tokens}}.
```
jricher, line 3426:
```suggestion
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -3321,25 +3327,31 @@ The AS validates that the token presented is associated with the management
 URL, that the AS issued the token to the given client instance, and that
 the presented key is appropriate to the token.
 
-If the access token has expired, the AS SHOULD honor the rotation request to
-the token management URL since it is likely that the client instance is attempting to
+Note that in many cases, the access token will have expired for regular use. To facilitate
+token rotation, the AS SHOULD honor the rotation request of the expired access token
+since it is likely that the client instance is attempting to
 refresh the expired token. To support this, the AS MAY apply different lifetimes for
 the use of the token in management vs. its use at an RS. An AS MUST NOT
```
jricher, line 3334:
```suggestion
refresh the expired token. To support this, the AS MAY allow a longer lifetime for
token management compared to its use at an RS. An AS MUST NOT
```

PR #378, opened 2022-02-10
provide more guidance on symmetric keys

### LINKED ISSUE #299: Why do we even support symmetric keys?
I don't see any reason to support symmetric keys for Client identification etc. Removing them will remove a (small) point of complexity, at zero cost.

Moreover, whenever people use "shared keys" nowadays these are usually passwords (even if we say explicitly not to use passwords as keys) and those come with a bunch of security issues.

In other words, is anybody still using shared symmetric keys in a justified **and secure** way?

fimbault:
Close #299 

- soften wording on quantum protection (may)
- add " The security of this protocol when in use with symmetric keys relies on the cryptographic security of the keys themselves. Therefore, a key management system MUST be used to supply the keys. Moreover, the symmetric key MUST NOT be a human memorable password or a value derived from one."

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4819,7 +4819,7 @@ access token by creating a valid signature. With asymmetric cryptography, the RS
 to know the public key associated with the token in order to validate, and therefore cannot
 create any new calls.
 
-Symmetric keys also have the expected advantage of providing better protection against quantum
+Symmetric keys may have the expected advantage of providing better protection against quantum
 threats in the future. Also, these types of keys (and their secure derivations) are widely
 supported among many cloud-based key management systems.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -4833,6 +4833,8 @@ not its value can be sent. This allows the AS to use pre-registered symmetric ke
 as key derivation schemes to take advantage of symmetric cryptography but without requiring
 key distribution at runtime, which would expose the keys in transit.
 
+The security of GNAP when in use with symmetric keys relies on the cryptographic security of the keys themselves. Therefore, a key management system MUST be used to supply the keys. Moreover, the symmetric key MUST NOT be a human memorable password or a value derived from one.
```
jricher, line 4836: Security considerations shouldn't have normative requirements. This text is good, but it needs to be up in the main body. Maybe in the section on key proofing and key material?

jricher: Target a new section 7.1.2 for this text.

fimbault:
Umm, again the editor config. For simplicity I'm just going to remove that PR and recreate it correctly.

PR #379, opened 2022-02-10
assertions as array

### LINKED ISSUE #41: Assertion format names
§2.2 Requesting User Information: Editor's note: 

The `assertions` values are lifted from [[RFC8693](https://www.rfc-editor.org/info/rfc8693)]'s "token type identifiers" list, but is there a better source?

§2.4: Identifying the User: Editor's note:

The `assertions` keys are lifted from [RFC8693]'s "token type identifiers" list, but is there a better source? Additionally: should this be an array of objects with internal typing like the sub_ids? Do we expect more than one assertion per user anyway?

§3.4 Returning User Information: Editor's note:

Should `assertion` be an array of objects with internal typing like the sub_ids? Do we expect more than one assertion per user anyway?

fimbault:
Closes #41 with array of objects to account for the possibility of several assertions

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -2337,7 +2337,6 @@ sub_ids (array of objects)
     `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. The assertion
     values are the string serialization of the assertion format, encoded as a plain
     JSON string. Additional assertion types are defined by [a registry TBD](#IANA).
-    \[\[ [See issue #41](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/41) \]\]
 
 updated_at (string)
 : Timestamp as an ISO8610 date string, indicating
@@ -2354,7 +2353,7 @@ updated_at (string)
     "id": "XUT2MFM1XBIKJKSDU8QM"
   } ],
   "assertions": [ {
-    "format": "id_token",
+    "type": "id_token",
     "value": "eyj..."
   } ]
 }
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -988,8 +988,8 @@ A non-normative example of a grant request is below:
         }
     },
     "subject": {
-        "sub_ids_formats": ["iss_sub", "opaque"],
-        "assertions_formats": ["id_token"]
+        "sub_id_formats": ["iss_sub", "opaque"],
+        "assertion_formats": ["id_token"]
     }
 }
 ~~~
@@ -1182,11 +1182,11 @@ the AS, it sends a `subject` field as a JSON object. This object MAY
 contain the following fields (or additional fields defined in
 [a registry TBD](#IANA)).
 
-sub_ids_formats (array of strings)
+sub_id_formats (array of strings)
 : An array of subject identifier subject formats
             requested for the RO, as defined by {{I-D.ietf-secevent-subject-identifiers}}.
 
-assertions_formats (array of strings)
+assertion_formats (array of strings)
 : An array of requested assertion formats. Possible values include
     `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. Additional
     assertion formats are defined by [a registry TBD](#IANA).
@@ -1194,8 +1194,8 @@ assertions_formats (array of strings)
 
 ~~~
 "subject": {
-  "sub_ids_formats": [ "iss_sub", "opaque" ],
-  "assertions_formats": [ "id_token", "saml2" ]
+  "sub_id_formats": [ "iss_sub", "opaque" ],
+  "assertion_formats": [ "id_token", "saml2" ]
 }
 ~~~
 
@@ -1211,7 +1211,7 @@ channels by the client instance, as discussed in {{response-subject}}.
 
 The AS SHOULD NOT re-use subject identifiers for multiple different ROs.
 
-Note: the "sub_ids_formats" and "assertions_formats" request fields are independent of
+Note: the "sub_id_formats" and "assertion_formats" request fields are independent of
 each other, and a returned assertion MAY use a different subject
 identifier.
 
@@ -2329,11 +2329,10 @@ sub_ids (array of objects)
             RO, as defined by
             {{I-D.ietf-secevent-subject-identifiers}}.
 
-: An array containing assertions as values keyed on the assertion
-    type defined by [a registry TBD](#IANA). Possible keys include
-    `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. The assertion
-    values are the string serialization of the assertion format, encoded as a plain
-    JSON string. Additional assertion types are defined by [a registry TBD](#IANA).
+: An array containing assertions as objects each containing the assertion
+    format and the assertion value as the JSON string serialization of the assertion. 
+    Possible formats include `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. 
+    Additional assertion formats are defined by [a registry TBD](#IANA).
 
 updated_at (string)
 : Timestamp as an ISO8610 date string, indicating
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -1394,8 +1394,8 @@ sub_ids (array of objects)
 : An array of subject identifiers for the
             end user, as defined by {{I-D.ietf-secevent-subject-identifiers}}.
 
-assertions (object)
-: An object containing assertions as values keyed on the assertion
+assertions (array of objects/strings)
```
jricher, line 1397:
```suggestion
assertions (array of objects)
```

I don't see anything here about string-formatted assertions, so should this just be objects?

jricher (changes requested):
I think this might need a bit more thought if we're going to change the syntax of it. The old single-object structure was meant to allow for a single ID token, a single SAML assertion, etc. As an array it might make more sense to have the object have an internal "type" field instead, or "format" to match the subject identifier.

adeinega:
As a completely random comment, I find it interesting that computer science commonly refers to an array as a collection of elements with the same data type. It's also a fixed-size collection of elements. https://datatracker.ietf.org/doc/html/rfc8259#section-5 clearly defines and uses an array. Moreover, it says that

> There is no requirement that the values in an array be of the same type.

what's a bit more important is that this specification should probably use "array of ..." everywhere as this isn't true right now. I see the use of both "list of ..." and "array of ..." at the same time.

jricher:
Since GNAP is based around JSON for its data model, we use JSON-native data types and their names. The name for an unlabeled, ordered collection is a "JSON array", as you pointed out. If there are places in GNAP where we're using "list" where we should be saying "array", we should fix that.

draft-ietf-gnap-core-protocol.md
fimbault: ok

fimbault:
Todo : 
- inner description to change (key)
- assertions -> format
- subid_formats or something like that

draft-ietf-gnap-core-protocol.md
```diff
@@ -2332,12 +2329,11 @@ sub_ids (array of objects)
             RO, as defined by
             {{I-D.ietf-secevent-subject-identifiers}}.
 
-: An object containing assertions as values keyed on the assertion
+: An array containing assertions as values keyed on the assertion
     type defined by [a registry TBD](#IANA). Possible keys include
     `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. The assertion
     values are the string serialization of the assertion format, encoded as a plain
     JSON string. Additional assertion types are defined by [a registry TBD](#IANA).
```
jricher, line 2336:
```suggestion
: An array containing assertions as objects each containing the assertion
    format and the assertion value as the JSON string serialization of the assertion. 
    Possible formats include `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. 
    Additional assertion formats are defined by [a registry TBD](#IANA).
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4540,9 +4537,9 @@ key_proofs_supported (array of strings)
...
           [key section](#key-format) of the request.
 
-subject_formats_supported (array of strings)
+sub_ids_formats_supported (array of strings)
 : OPTIONAL. A list of the AS's supported
-          subject identifier types. The values of this list correspond to possible values
+          subject identifier formats. The values of this list correspond to possible values
           of the [subject identifier section](#request-subject) of the request.
 
 assertions_supported (array of strings)
```
jricher, line 4545:
```suggestion
assertion_formats_supported (array of strings)
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4540,9 +4537,9 @@ key_proofs_supported (array of strings)
           values of the `proof` field of the
           [key section](#key-format) of the request.
 
-subject_formats_supported (array of strings)
+sub_ids_formats_supported (array of strings)
```
jricher, line 4540:
```suggestion
sub_id_formats_supported (array of strings)
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1182,20 +1182,20 @@ the AS, it sends a `subject` field as a JSON object. This object MAY
...
     `id_token` for an {{OIDC}} ID Token and `saml2` for a SAML 2 assertion. Additional
-    assertion values are defined by [a registry TBD](#IANA).
+    assertion formats are defined by [a registry TBD](#IANA).
     \[\[ [See issue #41](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/41) \]\]
 
 ~~~
 "subject": {
-  "formats": [ "iss_sub", "opaque" ],
-  "assertions": [ "id_token", "saml2" ]
+  "sub_ids_formats": [ "iss_sub", "opaque" ],
```
jricher, line 1197:
```suggestion
  "sub_id_formats": [ "iss_sub", "opaque" ],
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1182,20 +1182,20 @@ the AS, it sends a `subject` field as a JSON object. This object MAY
 contain the following fields (or additional fields defined in
 [a registry TBD](#IANA)).
 
-formats (array of strings)
-: An array of subject identifier subject types
+sub_ids_formats (array of strings)
```
jricher, line 1185:
```suggestion
sub_id_formats (array of strings)
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -988,8 +988,8 @@ A non-normative example of a grant request is below:
         }
     },
     "subject": {
-        "formats": ["iss_sub", "opaque"],
-        "assertions": ["id_token"]
+        "sub_ids_formats": ["iss_sub", "opaque"],
```
jricher, line 991:
```suggestion
        "sub_id_formats": ["iss_sub", "opaque"],
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -988,8 +988,8 @@ A non-normative example of a grant request is below:
...
+        "assertions_formats": ["id_token"]
```
jricher, line 992:
```suggestion
        "assertion_formats": ["id_token"]
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1182,20 +1182,20 @@ the AS, it sends a `subject` field as a JSON object. This object MAY
...
 [a registry TBD](#IANA)).
 
-formats (array of strings)
-: An array of subject identifier subject types
+sub_ids_formats (array of strings)
+: An array of subject identifier subject formats
             requested for the RO, as defined by {{I-D.ietf-secevent-subject-identifiers}}.
 
-assertions (array of strings)
+assertions_formats (array of strings)
```
jricher, line 1189:
```suggestion
assertion_formats (array of strings)
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1182,20 +1182,20 @@ the AS, it sends a `subject` field as a JSON object. This object MAY
...
-    assertion values are defined by [a registry TBD](#IANA).
+    assertion formats are defined by [a registry TBD](#IANA).
     \[\[ [See issue #41](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/41) \]\]
 
 ~~~
 "subject": {
-  "formats": [ "iss_sub", "opaque" ],
-  "assertions": [ "id_token", "saml2" ]
+  "sub_ids_formats": [ "iss_sub", "opaque" ],
+  "assertions_formats": [ "id_token", "saml2" ]
```
jricher, line 1198:
```suggestion
  "assertion_formats": [ "id_token", "saml2" ]
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -1211,7 +1211,7 @@ channels by the client instance, as discussed in {{response-subject}}.
 
 The AS SHOULD NOT re-use subject identifiers for multiple different ROs.
 
-Note: the "formats" and "assertions" request fields are independent of
+Note: the "sub_ids_formats" and "assertions_formats" request fields are independent of
```
jricher, line 1214:
```suggestion
Note: the "sub_id_formats" and "assertion_formats" request fields are independent of
```

PR #380, opened 2022-02-10
data URI for logo

### LINKED ISSUE #48: Logo by value
§2.3.3 Providing Displayable RC Information

Editor's note: 

Would we want to support pushing a display logo by value? On the upside it allows for more dynamic detached clients and doesn't require the AS to fetch information. On the downside, this is harder for the AS to enforce a policy about and could lead to potential exploits caused by sending binary image files.

fimbault:
Close #48 

Should we add a security consideration ?

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -44,6 +44,7 @@ normative:
         -
           ins: P. Saint-Andre
     RFC2119:
+    RFC2397:
     RFC3986:
     RFC5646:
     RFC7231:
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -1337,17 +1338,19 @@ uri (string)
...
 ~~~
 "display": {
     "name": "My Client Display Name",
-    "uri": "https://example.net/client"
+    "uri": "https://example.net/client", 
+    "logo_uri": "data:image/png;base64,Eeww...="
 }
 ~~~
 
+The logo MAY be passed by value by using a data: URI {{!RFC2397}} referencing an image mediatype.
```
adeinega, line 1352: I suggest moving this sentence to the description of "logo_uri", I don't think you want someone to think it's allowed for other places like "uri" and so forth.

fimbault: yes we can do that

russhousley:
There is a lot of information about fetching or embedding logo images in draft-ietf-lamps-rfc3709bis.

What are the concerns if the web server returns a different image than the one that was originally configured?  This is prevented in the referenced draft by providing a URL and a hash of the image.

What are the privacy concerns associated with the web servers seeing who and when the image file is fetched?  Embedding resolves this concern, but the document says that one MAY embed the image.

What image formats are allowed?  I think you should pick a mandatory-to-implement, but MAY support others too.

jricher:
@russhousley there's a section on SSRF attacks for fetched URIs and another on trusting and exposing client-provided URIs, of which the logo URI is a main example. If these sections can be expanded, please open issues and suggest text for them. Thank you!

Issue #353, opened 2021-11-19
Modifying a request during pending interaction


pq2:
[Section 5.3](https://www.ietf.org/archive/id/draft-ietf-gnap-core-protocol-08.html#name-modifying-an-existing-reque) explains how an existing request can be modified by a PATCH request to the continuation endpoint. My question is, what happens if such a PATCH request is sent while the interaction for the associated grant request is not yet finished? I feel like this should not be allowed, but could not find any information anywhere in Section 5 that this is actually forbidden.

Section 5.3 also says that the PATCH request may include _post-interaction responses_. I assume this means that by specifying an interaction reference in the PATCH request the interaction can be completed by the PATCH request, so the question would be what happens if the interaction is not completed and the PATCH request does not contain an interaction reference?

pq2:
In Section 3.3.4 it says 
`If the AS returns a nonce, the client instance MUST NOT continue a grant request before it receives the associated interaction reference on the callback URI.`
If the PATCH request is considered a continuation request (which is reasonable since it is sent to the continuation URI), this would prohibit a PATCH request before the interaction finishes. However, if no interaction finish mode is used, this does not apply.

fimbault:
I believe the cases are covered, even if we could make things clearer probably : 
- using PATCH instead of POST aims to provide better signal/control of when the partial update can be made as part of a continuation request (and it needs to use the continue access token)
- when the client can't receive a finish method (redirect or push), then it will need to poll regularly (cf 5.2)

jricher:
Summary from editor's discussion:

The grant process is stateful, a grant goes from (requested) -> (pending approval) -> (approved), at least. The PATCH request only makes sense to an (approved) grant, the post-interaction requests only make sense for the (pending) grant. We can call out this stateful nature and constrain different requests and inputs to specific states.

jricher:
Proposed solution to make a state diagram and add state description in the overall protocol discussion.

PR #382, opened 2022-02-10
one time use clarification

### LINKED ISSUE #97: Interaction requests as one-time-use
§5.3 Modifying an Existing Request: Editor's note: 

> Since the original callback was already used in the initial exchange, and the callback is intended for one-time-use, a new one needs to be included in order to use the callback again.

The net result of this is that interaction requests are really only meant to be responded to exactly once by the AS. This isn't spelled out explicitly, but could be included in Section 2.5 and/or Section 3.3.

fimbault:
Close #97


jricher:
Still need to explain better how to do that, will track as part of lifecycle discussion #353

PR #384, opened 2022-02-10
clarify token management lifecycle

### LINKED ISSUE #161: Terminology & lifecycle: revoke / rotate

fimbault:
Close #161 

Actually a very minor editorial change to specify the actions.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3286,7 +3286,7 @@ The AS SHOULD revoke all associated access tokens.
 If an access token response includes the `manage` parameter as
 described in {{response-token-single}}, the client instance MAY call
 this URL to manage the access token with any of the actions defined in
-the following sections. Other actions are undefined by this
+the following sections: rotate and revoke. Other actions are undefined by this
 specification.
 
 The access token being managed acts as the access element for its own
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3286,15 +3286,15 @@ The AS SHOULD revoke all associated access tokens.
...
+the following sections: rotate and revoke. Other actions are undefined by this
 specification.
 
 The access token being managed acts as the access element for its own
 management API. The client instance MUST present proof of an appropriate key
 along with the access token.
 
 If the token is sender-constrained (i.e., not a bearer token), it
-MUST be sent [with the appropriate binding for the access token](#use-access-token).
+MUST be sent [with the appropriate binding for the access token](#use-access-token).
```
jricher, line 3297:
```suggestion
MUST be sent [with the appropriate binding for the access token](#use-access-token).
```

PR #385, opened 2022-02-10
expand error codes

aaronpk:
for #79


draft-ietf-gnap-core-protocol.md
```diff
@@ -2428,33 +2428,39 @@ If the AS determines that the request cannot be issued for any
...
-~~~
-{
-
-  "error": "user_denied"
-
-}
-~~~
+     "invalid_client":
+     :     The request was made from a client that was not recognized
+           or allowed by the AS, or the client's signaure validation failed.
```
fimbault, line 2441: small typo : signature

draft-ietf-gnap-core-protocol.md
```diff
@@ -2428,33 +2428,38 @@ If the AS determines that the request cannot be issued for any
...
 
-too_fast
-: The client instance did not respect the timeout in the
-            wait response.
+  "error": "user_denied"
 
-unknown_request
-: The request referenced an unknown ongoing access request.
+}
+~~~
```
yaronf, line 2462: I suggest to add:

> Per policy, highly secure AS implementations MAY elect to return `user_denied` upon any error case, in order not to leak any information.

jricher: I personally don't like this suggestion. `user_denied` should mean that specifically the RO denied the request. If you want to hide the reasons for things failing, there should be a generic `unknown_error` as an option. (Which I'm also not sure about, but like it more than co-opting an otherwise meaningful error code.)

jricher: @yaronf "access denied" is not the same as "user denied". "User denied" specifically means that the user interactively said "no" to a decision.

yaronf: OK, so maybe we also need a more generic `denied` error code, which means "denied for an unspecified reason".

PR #386, opened 2022-02-14
Use "array of"

adeinega:
For the sake of consistency.

See https://github.com/ietf-wg-gnap/gnap-core-protocol/pull/379#issuecomment-1039615442 for an initial discussion on that.


jricher:
Looks good, thank you!

PR #388, opened 2022-02-16
Strengthened normative requirements for continue timeout

### LINKED ISSUE #86: Default polling wait time
§5 Continuing a Grant Request: Editor's note: 

What's a reasonable amount of time to wait by default so as not to DOS the server if the field is missing?

jricher:
Closes #86﻿

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1878,7 +1878,7 @@ wait (integer)
 : RECOMMENDED. The amount of time in integer
             seconds the client instance MUST wait after receiving this continuation
             handle and calling the URI. The value SHOULD NOT be less than five seconds,
-            and omission of the value MUST NOT be interpreted as zero (i.e., no delay 
+            and omission of the value MUST NOT be interpreted as zero (i.e., no delay
             between requests)
 
 access_token (object)
@@ -2887,7 +2887,7 @@ Digest: sha256=...
 
 If a `wait` parameter was included in the [continuation response](#response-continue), the
 client instance MUST NOT call the continuation URI prior to waiting the number of
-seconds indicated. If no `wait` period is indicated, the client instance 
+seconds indicated. If no `wait` period is indicated, the client instance
 MUST NOT poll immediately and SHOULD
 wait at least 5 seconds. If the client instance does not respect the
 given wait period, the AS MUST return the error `too_fast` defined in {{response-error}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -1876,8 +1876,10 @@ uri (string)
 
 wait (integer)
 : RECOMMENDED. The amount of time in integer
-            seconds the client instance SHOULD wait after receiving this continuation
-            handle and calling the URI.
+            seconds the client instance MUST wait after receiving this continuation
+            handle and calling the URI. The value SHOULD NOT be less than five seconds,
+            and omission of the value MUST NOT be interpreted as zero (i.e., no delay
+            between requests)
```
jricher, line 1882:
```suggestion
            between requests).
```

adeinega:
As a bit random comment... I always wondered if is it allowed for the AS to provide a "delayed" response, or in other words, leverage the HTTP long polling mechanism just like for this response with the `continue` field.

Of course, there will be definitely upsides and downsides to it.

See https://datatracker.ietf.org/doc/html/rfc6202 for more info.

draft-ietf-gnap-core-protocol.md
```diff
@@ -1876,8 +1876,10 @@ uri (string)
 
 wait (integer)
 : RECOMMENDED. The amount of time in integer
-            seconds the client instance SHOULD wait after receiving this continuation
-            handle and calling the URI.
+            seconds the client instance MUST wait after receiving this continuation
```
adeinega, line 1879: I suggest to omit "seconds" and use `The amount of time in seconds...`.

Another minor comment is that JSON doesn't have the integer datatype, it has the "number" datatype.

jricher: The value is an integer even if the holder is a float. We don't want to encourage fractional second timeouts here.

jricher:
Since it's over HTTP, long polling is allowed regardless of whether we mention it or not. I personally don't see a reason to mention it explicitly here.

draft-ietf-gnap-core-protocol.md
adeinega: What you get right now is 

> The amount of time in integer seconds

and that sounds awkward to me.

I understand that you don't want to encourage fractional second timeuts. I just wanted to say is that there is no such data type as "integer" in JSON, there is a JSON number, it might be a minor (but important) detail if you want to be precise as much as possible, and use JSON data types.

jricher: Right - the value is in **integer seconds**, and it is encoded as a **JSON number**. It reads fine to me, I've seen the construct "integer seconds" used plenty of times in similar works.

PR #389, opened 2022-02-16
Redefine user code interaction methods

### LINKED ISSUE #9: Clarify User-Code Interaction
Yaron: it is unclear to me what the goal of the user-code interaction is, in other words what is the threat model. If it's just "second factor, something you have", then an OTP MFA device would be simpler and offer the same security.

Justin: This is an equivalent function of the OAuth Device Flow: https://tools.ietf.org/html/rfc8628

It’s not about a second factor for authenticating the user, really, it’s about facilitating authorization on a secondary device through the use of a user-typed code to tie the back end and front end together.

Tom Jones: Ok - several others have been asking as well.  I wrote up this page https://tcwiki.azurewebsites.net/index.php?title=Apps_on_User_Devices

basically there are assertions from the programmer of the app, and from the app itself that are passed in the id token (whatever) from the user device.
Note that this means that not all assertions are signed by the same key in this case!
Nor are all the attributes about the subject. Some are just in support of the level of assurance.
The big difference with FIDO U2F and II is that the keys are bound to the user in the device, not in the RP.

### LINKED ISSUE #123: Stable URL for “Display of a Short User Code”
In the section “Display of a Short User Code” it says the URL *MUST* be stable so that the client can be statically configured with it. This seems to contradict the idea of returning the URL in the response here especially since the URL isn’t even required. Either the client is statically configured with a URL and so it doesn’t need the URL in the response, or the client needs the URL in the response because it is not statically configured. It seems the MUST is too restrictive here.

jricher:
Splits user code into two methods, `user_code` with no URI in the response and `user_code_uri` with a dynamic URI in the response.

Closes #123﻿
Closes #9

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -2645,9 +2645,8 @@ the client instance has to communicate the redirection URI to the end user.
 
 In many cases, the URI indicates a web page hosted at the AS, allowing the
 AS to authenticate the end user as the RO and interactively provide consent.
-If the URI is hosted by the AS,
-the AS MUST determine the grant request being referenced from the URL
-value itself. If the URL cannot be associated with a currently active
+The URI value is used to identify the grant request being authorized.
+If the URL cannot be associated with a currently active
 request, the AS MUST display an error to the RO and MUST NOT attempt
 to redirect the RO back to any client instance even if a [redirect finish method is supplied](#request-interact-callback-redirect).
 If the URI is not hosted by the AS directly, the means of communication between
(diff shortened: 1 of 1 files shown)


fimbault:
Seems fine, interesting PR

draft-ietf-gnap-core-protocol.md
```diff
@@ -2236,56 +2267,99 @@ object that contains the following members.
 
 code (string)
 : REQUIRED. A unique short code that the user
-              can type into an authorization server. This string MUST be
+              can type into a static URL. This string MUST be
```
aaronpk, line 2270:
```suggestion
              can type into a web page. This string MUST be
```

PR #394, opened 2022-02-20
Minor edits

yaronf:


jricher (approved):

draft-ietf-gnap-core-protocol.md
```diff
@@ -218,7 +218,7 @@ Resource Server (RS)
 Resource Owner (RO)
 : subject entity that may grant or deny operations on resources it has authority upon.
 
-    Note: the act of granting or denying an operation may be manual (i.e. through an interaction with a physical person) or automatic (i.e. through predefined organizational rules).
+    Note: the act of granting or denying an operation may be manual (i.e. through an interaction with a person) or automatic (i.e. through predefined organizational rules).
```
jricher, line 221: This is the only one I'm not sure of -- @fimbault do you remember if the "physical person" aspect was specifically needed in this definition?

fimbault: I guess that's because there can exist a "moral person", so we actually meant a human here

jricher: Suggestion to revert the change: 

```suggestion
    Note: the act of granting or denying an operation may be manual (i.e. through an interaction with a physical person) or automatic (i.e. through predefined organizational rules).
```

PR #399, opened 2022-02-23
new 7.1.2 section on key protection

### LINKED ISSUE #299: Why do we even support symmetric keys?

fimbault:
Close #299

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3525,6 +3525,10 @@ by both clients and ASs, which could be accomplished by e.g. a client publishing
 a public key at a URL. For interoperability, this method could later be described
 as an extension.
 
+### Key Protection {#key-protection}
+
+The security of GNAP when in use with symmetric keys relies on the cryptographic security of the keys themselves. 
+Therefore, a key management system MUST be used to supply the keys. Moreover, the symmetric key MUST NOT be a human memorable password or a value derived from one.
 
 ## Presenting Access Tokens {#use-access-token}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3525,6 +3525,10 @@ by both clients and ASs, which could be accomplished by e.g. a client publishing
 a public key at a URL. For interoperability, this method could later be described
 as an extension.
 
+### Key Protection {#key-protection}
+
+The security of GNAP when in use with symmetric keys relies on the cryptographic security of the keys themselves. 
+Therefore, a key management system MUST be used to supply the keys. Moreover, the symmetric key MUST NOT be a human memorable password or a value derived from one.
```
jricher, line 3531: I like what you're doing here, but I think maybe we can do something like this instead:

```suggestion
The security of GNAP relies on the cryptographic security of the keys themselves. 
When used with symmetric keys, a key management system or secure key derivation mechanism MUST be used to supply the keys. Symmetric keys MUST NOT be a human memorable password or a value derived from one.
```

russhousley:
Please take a look a BCP 107 (a.k.a. RFC 4107).  You may need a reference here.

PR #415, opened 2022-03-20
add diagram for session attack

jricher:


draft-ietf-gnap-core-protocol.md
```diff
@@ -5419,6 +5419,68 @@ use polling to continue the request. The tradeoffs of this approach are discusse
 {{security-polling}}, and if possible, an explicit interaction finish method should be
 used instead.
 
+## Calculating Interaction Hash {#security-interact-hash}
+
+The calculation of the interaction hash value provides defence in depth, allowing a client
+instance to protect itself from spurious injection of interaction references when using a
+callback finish method. An attacker attempting to substitute the interaction reference
+is stopped in several places.
```
aaronpk, line 5427: I think it'd be worth calling out the main way injection is prevented (by the AS) here as well.

jricher, line 5427:
```suggestion
callback finish method. The AS is protected during this attack through the
continuation access token being bound to the expected interaction reference,
but without hash calculation, the attacker could cause the client to make an
HTTP request on command. With both of these in place, an attacker attempting to substitute the interaction reference
is stopped in several places.
```

PR #421, opened 2022-05-04
Expanded key proofing field to allow parameters

### LINKED ISSUE #417: Key Proofing Parameters
Some key proofing methods have additional options that could be signaled in the GNAP protocol structure. Notable, HTTP Message Signatures has the ability to use different HTTP Signing Algorithms and different HTTP Digest Algorithms.

The `proof` field could be changed to an object to accomodate this kind of use, allowing each proofing method to define its own parameters:

```
{
  "proof": {
    "method": "httpsig",
    "alg": "rsa-pss-sha512",
    "digest-alg": "sha-512"
   }
}
```

Each method can also define "Default" values for missing parameters, allowing this to collapse in the simple case back to a string as it is today:

```
{
  "proof": "httpsig"
}
```

These string values would have a deterministic expansion defined in the core protocol.

jricher:
Expands the key `proof` parameter to allow it to be an object or a string, to facilitate methods with multiple parameters such as HTTPSig.

Closes #417


draft-ietf-gnap-core-protocol.md
```diff
@@ -3733,9 +3740,40 @@ Values defined by this specification are as follows:
...
+
+This compact form is equivalent to sending the following object:
+
+~~~ json
+{
+    "proof": {
+        "method": "httpsig"
+    }
+}
+~~~
```
jricher, line 3775:
```suggestion
as a string instead of an object. For example, the `mtls` method with no additional parameters could be sent instead as:

~~~ json
{
    "proof": "mtls"
}
~~~

This compact form is equivalent to sending the following object:

~~~ json
{
    "proof": {
        "method": "mtls"
    }
}
~~~
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -3811,9 +3849,17 @@ NOTE: '\' line wrapping per RFC 8792
...
+Message Signature as described in {{I-D.ietf-httpbis-message-signatures}}. This method defines the following parameters:
+
+`alg`
+: The explicit HTTP signature algorithm, from the HTTP Signature Algorithm registry. If
+    this parameter is omitted, the signing algorithm MUST be derived from the key
+    material (such as using the JWS algorithm in a JWK formatted key). OPTIONAL.
+
+`content-digest-alg`
+: The algorithm used for the Content-Digest field, used to protect the body. If this
+    parameter is omitted, its value is `sha-256`. OPTIONAL.
```
jricher, line 3862:
```suggestion
`alg`:
: The explicit HTTP signature algorithm, from the HTTP Signature Algorithm registry. If
    this parameter is omitted, the signing algorithm MUST be derived from the key
    material (such as using the JWS algorithm in a JWK formatted key). OPTIONAL.

`content-digest-alg`:
: The algorithm used for the Content-Digest field, used to protect the body. If this
    parameter is omitted, its value is `sha-256`. OPTIONAL.
```

aaronpk (approved):

PR #423, opened 2022-05-16
Access rights as strings/objects

### LINKED ISSUE #407: Access rights by reference
We say: 

> Access rights MAY be communicated as a string known to the AS or RS representing the access being requested. Each string SHOULD correspond to a specific expanded object representation at the AS.

But then we give examples like "some other thing" that clearly do not correspond to anything. We also don't specify any kind of semantics, such as if I request the reference string, I might get back the expanded access right representation.

I suggest to remove or weaken that sentence: 

> Each string may or may not correspond to a specific expanded object representation at the AS.

jricher:
Removes the normative suggestion that access right string values be always expandable to objects, but instead adds an example of one way to do it.

Closes #407


draft-ietf-gnap-core-protocol.md
```diff
@@ -4448,6 +4474,11 @@ as well as the reference values of `read`, `dolphin-metadata`, and `some other t
 The requested access is the union of all elements of the array, including both objects and
 reference strings.
 
+
+
+In order to facilitate the use of both object and reference strings to access the same
+kind of APIs, the API designer
+
```
jricher, line 4481:
```suggestion
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4403,6 +4404,31 @@ API designer choosing any such human-readable strings SHOULD take steps
 to ensure the string values are not easily confused by a developer,
 such as by limiting the strings to easily disambiguated characters.
 
+One possible approach for choosing reference string values is to use the same value as the
+`type` parameter from the fully-specified object, with the API defining a set of default
+behaviors in this case. For example, an API definition could declare the following string:
+
```
jricher, line 4410:
```suggestion
In order to facilitate the use of both object and reference strings to access the same
kind of APIs, the API designer can define a clear mapping between these forms.
One possible approach for choosing reference string values is to use the same value as the
`type` parameter from the fully-specified object, with the API defining a set of default
behaviors in this case. For example, an API definition could declare the following string:

```

aaronpk (approved):

PR #426, opened 2022-06-29
RO Policy

### LINKED ISSUE #425: Further clarifying End-user / RO
Following the discussion on the mailing list, will plan a rewording to clarify
https://mailarchive.ietf.org/arch/msg/txauth/URII4ZuCJsYXC9m3_GrWsuFDQz4/

fimbault:
to close #425: update section 4 to provide an additional use case (RO's policy).

I don't think we need to update the definitions or roles.md

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -2391,7 +2391,7 @@ need to sign off on a transfer request. Alternatively, the AS could require N of
 to approve a given request in order. The AS could also determine that the end user is not the appropriate RO
 for a given request and reach out to the appropriate RO asynchronously.
 
-The RO is also allowed to define a policy at the AS to determine which kind of end user can get access to the resource, and under which condition. For instance, such a condition might require the end user login and the acceptance of the RO's legal provisions.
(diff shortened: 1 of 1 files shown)


jricher:
@fimbault please run `eclint fix` on this PR to clean up the line end spaces, otherwise looks good to me. Thanks!

fimbault:
will do !

draft-ietf-gnap-core-protocol.md
```diff
@@ -2389,8 +2389,11 @@ The AS is also allowed to request authorization from more than one RO, if the AS
 record might need to be released by both an attending nurse and a physician, or both owners of a bank account
 need to sign off on a transfer request. Alternatively, the AS could require N of M possible RO's
 to approve a given request in order. The AS could also determine that the end user is not the appropriate RO
-for a given request and reach out to the appropriate RO asynchronously. The details of determining which RO's are required for a given
-request are out of scope for this specification.
+for a given request and reach out to the appropriate RO asynchronously.
+
+The RO is also allowed to define a policy at the AS to determine which kind of end user can get access to the resource, and under which condition. For instance, such a condition might require the end user login and the acceptance of the RO's legal provisions.
```
jricher, line 2394:
```suggestion
The RO is also allowed to define a policy at the AS to determine which kind of end user can get access to the resource, and under which condition. For instance, such a condition might require the end user login and the acceptance of the RO's legal provisions. Alternatively, client software could be acting without an end user, and the RO's policy allows issuance of access tokens to instances of that client software without runtime approval.
```

PR #427, opened 2022-06-29
Added state diagram and discussion

### LINKED ISSUE #90: Contents of follow-in continue response
§5.2 Continuing During Pending Interaction:  Editor's note: 

Do we want to be more precise about what's expected inside the "continue" object? I think that at least the URI is required, access token required IF used, etc. This is even if they haven't changed since last time, and the client will use whatever value comes back?

### LINKED ISSUE #92: Enforcing modification of a request
§5.2 Modifying an Existing Request: Editor's note: 

We could state something like "resources and subject MUST NOT be the same as in the initial or previous request" to enforce that this really is a change, but is there value in calling that out here? Somehow we do probably want to tell the AS to not let a client simply post the same request here to rotate access tokens now that we've got an explicit function for that, right?

### LINKED ISSUE #93: Inclusion of the "user" field in a modification
§5.3 Modifying an Existing Request: Editor's note: 

This would allow the client to do things like gather the user's identifiers post-request, or gather an assertion from an on-device element that the AS can verify. It opens up potential avenues for trouble if the user here is different from the RO that's already showed up at the AS or race conditions if the RQ's identity changes mid-stream. But that said, this seems important for multi-log-in cases and the like, probably.

### LINKED ISSUE #94: Exclusion of "client" field in a modification
§5.3 Modifying an Existing Request: Editor's note: 

We do not want the client to be able to get swapped out from underneath the user, especially post-consent. However, including this field in a PATCH update request might be the place to define key rotation for the grant request itself, but we'd need to be very careful of how that works. And it feels like it might have consequences outside of the request, such as rotating the key for all ongoing grants for a given client instance, which isn't really desirable here. We need a lot more discussion and engineering on this before including it.

### LINKED ISSUE #95: Inclusion of post-interaction items in a modification
§5.3 Modifying an Existing Request: Editor's note: 

> The RC MAY include post-interaction responses such as described in Section 5.1. 

It seems a little odd to include this in a request but I can't see a reason to not allow it.

### LINKED ISSUE #353: Modifying a request during pending interaction

jricher:
Adds a diagram and discussion of protocol state.

Closes #353
Closes #95
Closes #94
Closes #93
Closes #92
Closes #90


yaronf (changes requested):

draft-ietf-gnap-core-protocol.md
```diff
@@ -307,6 +307,30 @@ A global assumption made by GNAP is that authorization requests are security and
...
+
+*Approved*
```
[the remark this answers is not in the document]
jricher: "Approved" is not a terminal state. Once something is approved it can be modified and later revoked.

"Rejected/denied" is an action, not a state, that puts the state back into "processing", where the AS determines what to do next. The AS could decide there's nothing else to do (the end user's decision is final) and go to "revoked", or it could move to "pending" if there's more that could happen to cause it to get approved, such as claims sent from the client software itself, or a separate interaction.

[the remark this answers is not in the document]
jricher: @yaronf I've added more text that hopefully makes this distinction more clear.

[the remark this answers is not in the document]
yaronf: @jricher I was referring to the *AS* making a decision to reject the request.

And now I'm even more confused. If both the AS and the RO are happy, what is the terminal state that denotes that the request was successful? It cannot be "revoked", can it.

draft-ietf-gnap-core-protocol.md
```diff
@@ -307,6 +307,30 @@ A global assumption made by GNAP is that authorization requests are security and
...
...an send a [polling continue request](#continue-poll) to the AS. This returns a [continue response](#response-continue) while the grant request remains in this state, allowing the client instance to continue to check the state of the pending grant request. If an [interaction finish method](#request-interact-finish) is specified in the grant request, the client instance can [continue the request after interaction](#continue-after-interaction) to the AS to move this request to the _processing_ state to be re-evaluated by the AS. Note that this occurs whether the grant request has been approved or denied by the RO, since the AS needs to take into account the full context of the request before determining the next step for the grant request. When other new information is available in the context of the grant request, such as through the asynchronous actions of the RO, the AS moves this request to the _processing_ state to be re-evaluated. If the AS determines that no additional interaction can occur, such as all the interaction methods have timed out or a [revocation request](#continue-delete) is received from the client instance, the grant request can be moved to the _finalized_ state.
```
yaronf, line 324:
```suggestion
: When a request needs to be approved by a RO, or interaction with the end user is required, the grant request enters a state of _pending_. In this state, no access tokens can be granted and no subject information can be released to the client instance. While a grant request is in this state, the AS seeks to gather the required [consent and authorization](#authorization) for the requested access. A grant request in this state is always associated with a _continuation access token_ bound to the client instance's key. If no [interaction finish method](#request-interact-finish) is associated with this request, the client instance can send a [polling continue request](#continue-poll) to the AS. This returns a [continue response](#response-continue) while the grant request remains in this state, allowing the client instance to continue to check the state of the pending grant request. If an [interaction finish method](#request-interact-finish) is specified in the grant request, the client instance can [continue the request after interaction](#continue-after-interaction) to the AS to move this request to the _processing_ state to be re-evaluated by the AS. Note that this occurs whether the grant request has been approved or denied by the RO, since the AS needs to take into account the full context of the request before determining the next step for the grant request. When new information is made available in the context of the grant request, such as through the asynchronous actions of the RO, the AS moves this request to the _processing_ state to be re-evaluated. If the AS determines that no additional interaction can occur, such as all the interaction methods have timed out or a [revocation request](#continue-delete) is received from the client instance, the grant request can be moved to the _finalized_ state.
```

Reworded "When new other".

draft-ietf-gnap-core-protocol.md
```diff
@@ -307,6 +307,30 @@ A global assumption made by GNAP is that authorization requests are security and
...
+
+*Finalized*
+: When an existing grant request is [revoked by the client instance](#continue-delete) or otherwise revoked by the AS (such as through out-of-band action by the RO), the grant request enters the _finalized_ state. This state can also be entered if the AS determines that no additional processing is possible, for example if the RO has denied the requested access or if interaction is required but no compatible interaction methods are available. This state can also be entered after the access tokens are issued, if the AS does not allow any additional processing on the grant request. Once in this state, no new access tokens can be issued, no subject information can be returned, and no interactions can take place. Once in this state, the grant request is dead and cannot be revived. If future access is desired by the client instance, a new grant request can be created, unrelated to this grant request.
```
yaronf, line 330: I suggest to rearrange this paragraph so that the "happy path" is first.

Also, I'm not clear what is meant by "if the AS does not allow any additional processing on the grant request". If this means "interaction", let's use this word instead of "processing".

draft-ietf-gnap-core-protocol.md
yaronf, line 324:
```suggestion
: When a request needs to be approved by a RO, or interaction with the end user is required, the grant request enters a state of _pending_. In this state, no access tokens can be granted and no subject information can be released to the client instance. While a grant request is in this state, the AS seeks to gather the required [consent and authorization](#authorization) for the requested access. A grant request in this state is always associated with a _continuation access token_ bound to the client instance's key. If no [interaction finish method](#request-interact-finish) is associated with this request, the client instance can send a [polling continue request](#continue-poll) to the AS. This returns a [continue response](#response-continue) while the grant request remains in this state, allowing the client instance to continue to check the state of the pending grant request. If an [interaction finish method](#request-interact-finish) is specified in the grant request, the client instance can [continue the request after interaction](#continue-after-interaction) to the AS to move this request to the _processing_ state to be re-evaluated by the AS. Note that this occurs whether the grant request has been approved or denied by the RO, since the AS needs to take into account the full context of the request before determining the next step for the grant request. When other information is made available in the context of the grant request, such as through the asynchronous actions of the RO, the AS moves this request to the _processing_ state to be re-evaluated. If the AS determines that no additional interaction can occur, such as all the interaction methods have timed out or a [revocation request](#continue-delete) is received from the client instance, the grant request can be moved to the _finalized_ state.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -2707,13 +2736,14 @@ NOTE: '\' line wrapping per RFC 8792
 
 # Continuing a Grant Request {#continue-request}
 
-While it is possible for the AS to return a [grant response](#response) with all the
+While it is possible for the AS to return an _approved_ [grant response](#response) with all the
```
yaronf, line 2739:
```suggestion
While it is possible for the AS to return an approved [grant response](#response) with all the
```

No need to italicize, we're not using "approved" as the name of the state.

yaronf (changes requested):

draft-ietf-gnap-core-protocol.md
jricher, line 330:
```suggestion
: After the access tokens are issued, if the AS does not allow any additional updates on the grant request, the grant request enters the _finalized_ state. This state is also entered when an existing grant request is [revoked by the client instance](#continue-delete) or otherwise revoked by the AS (such as through out-of-band action by the RO). This state can also be entered if the AS determines that no additional processing is possible, for example if the RO has denied the requested access or if interaction is required but no compatible interaction methods are available. Once in this state, no new access tokens can be issued, no subject information can be returned, and no interactions can take place. Once in this state, the grant request is dead and cannot be revived. If future access is desired by the client instance, a new grant request can be created, unrelated to this grant request.
```

yaronf:
The text (but probably not the diagram) should mention that each state can also transition into Finalized if a timeout occurs, e.g. if the client is expected to send a continuation request but doesn't do so.

jricher:
I think the current state text does mention the timeout in each state, but I did go back and add some more explicit explanation to #428 that was lacking.

PR #428, opened 2022-06-29
Allow interactions to time out

### LINKED ISSUE #416: Interaction Timeout
When an AS creates a set of of interaction artifacts, like a user code or interaction URI, these artifacts aren't likely to be available perpetually. The AS will likely want to purge these after some time.

It might be useful for the AS to indicate to the client a timeout for the "interact" return block so that the client has some idea of how much time it has to use the interaction methods.

```
{
    "interact": {
        "redirect": "https://server.example/interact/1234",
        "user_code": "123456",
        "expires_in": 600
    }
}
```

jricher:
Allows interaction responses from the AS to have an explicit timeout, with instructions on how the client instance can recover (do a modification).

Closes #416

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -5630,6 +5630,7 @@ Throughout many parts of GNAP, the parties pass shared references between each o
     - Expand proofing methods to allow definition by object, with single string as optimization for common cases.
     - Removed "split_token" functionality.
     - Collapse "user_code" into a string instead of an object.
+    - Allow interaction responses to time out.
 
 - -09
     - Added security considerations on redirection status codes.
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -2412,13 +2414,22 @@ If the AS and client instance have not reached a state where the delegation can
...
-To initiate an interaction start method indicated by the
+## Starting Interaction With the End User {#interaction-start}
+
+To initiate an interaction start modes indicated by the
 [interaction start responses](#response-interact) from the AS, the client instance
-follows the steps defined by that interaction method. The actions of the client instance
+follows the steps defined by that interaction start mode. The actions of the client instance
 required for the interaction start modes defined in this specification are described
-in the following sections.
+in the following sections. Interaction start modes defined in
```
yaronf, line 2427: Broken sentence

jricher: Thanks, missed that one on edit!

PR #432, opened 2022-07-04
Added title and link of informative source

pq2:
My master's thesis was published, so now I could fill in the source.


jricher:
Congratulations! Thanks for the reference.

PR #435, opened 2022-09-06
Key Rotation

jricher:
Adds a provision for rotating keys on access token requests. If a client needs to dynamically rotate its request key, this would happen through a client instance management 
API that is not defined in the core protocol.

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3333,10 +3333,12 @@ instance MUST present both the new key and previous key in the access token rota
 The client instance makes an HTTP POST as a JSON object with the following fields:
 
 `key`:
-: The new key value or reference in the format described in {{key-format}}.
+: The new key value or reference in the format described in {{key-format}}. Note that keys
+    passed by value are always public keys.
 
 `previous_key`:
-: The previous key value or by reference in format described in {{key-format}}.
+: The previous key value or by reference in format described in {{key-format}}. Note that keys
+    passed by value are always public keys.
 
 The `proof` method and parameters for the new key MUST be the same as those established for the
 previous key.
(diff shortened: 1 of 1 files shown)


fimbault (commented):
Great work. 

A few comments : 

- maybe we should have some kind of security consideration saying that we're expecting to send public keys
- small typo on signnature
- having the same parameters make sense in most cases (and i think it's reasonable), but is there a real limit as long as one provides the proof ?

fimbault (approved):

jricher:
Changing the parameters of the proof method really is the same level of function as changing the proof method entirely. Would you claim the parameters apply to the new key or the old key? And what if the proofing method makes it impossible to apply two signatures (for example) with different parameters? It opens up a can of worms to allow it to vary, and I think we need to cut all of that off.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3326,6 +3326,60 @@ NOTE: '\' line wrapping per RFC 8792
 
 \[\[ [See issue #103](https://github.com/ietf-wg-gnap/gnap-core-protocol/issues/103) \]\]
 
+### Binding a New Key to the Rotated Access Token {#rotate-access-token-key}
+
+If the client instance wishes to bind a new presentation key to an access token, the client
+instance MUST present both the new key and previous key in the access token rotation request.
+The client instance makes an HTTP POST as a JSON object with the following fields:
```
adeinega, line 3333: Nit: maybe `an HTTP POST request` for consistency reasons? The specification uses two forms `an HTTP POST request` and `an HTTP POST` at the same time.

adeinega:
I'm wondering if there are legit cases in the wild when as AS prohibits its clients to perform key rotation on their own. If so, how it will notify the clients?

adeinega:
@jricher, am I right that `previous_key` becomes immediately invalid? If it's true, wouldn't be nice to have here some way to rotate them more or less gracefully? It could be useful if there are multiple running instances of the client and it's tough, or even impossible, to update the keys on all these instances immediately (just as an example, the client instance could be deployed as a service in K8S).

Thank you.

PR #438, opened 2022-09-07
removed data models placeholder section

jricher:


fimbault:
I agree, maybe we could paste that into an issue of the RS draft (just as a reminder), since that's kind of related to the format of tokens etc

PR #441, opened 2022-09-28
Require Absolute URIs

### LINKED ISSUE #439: Are relative URLs allowed?
While it doesn't say so in the spec and the example suggest otherwise, are relative URLs allowed for properties such as the `continue` response.

Scenario, an implementation is deployed at a URL: server.example.com
The root of the server handles grant requests: `POST /` to `https://server.example.com`

The application code that builds the `continue` response must be able to construct the full URL and therefor have knowledge of the host at which the application is deployed including any path prefix.

This is not impossible, but in complex cloud deployments with load balancers and ingress services etc between the application code and the public endpoint this can get ugly.

The current example from Section 3.1 of the spec.
```
{
    "continue": {
        "access_token": {
            "value": "80UPRY5NM33OMUKMKSKU"
        },
        "uri": "https://server.example.com/continue",
        "wait": 60
    }
}
```
Implementations would be significantly easier to build if they could simply return:

```
{
    "continue": {
        "access_token": {
            "value": "80UPRY5NM33OMUKMKSKU"
        },
        "uri": "./continue",
        "wait": 60
    }
}
```

jricher:
Requires all URIs to be absolute throughout the protocol.

Closes #439

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1149,10 +1149,10 @@ to present to the RO during any interactive sequences.
 : Display name of the client software. RECOMMENDED.
 
 `uri` (string):
-: User-facing web page of the client software. This MUST be an absolute URI. OPTIONAL.
+: User-facing web page of the client software. This URI MUST be an absolute URI. OPTIONAL.
 
 `logo_uri` (string)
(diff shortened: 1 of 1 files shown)


fimbault (approved):
Makes sense

PR #443, opened 2022-10-05
Make keys single format only and add security considerations on equivalent keys

### LINKED ISSUE #18: RC identified by Multiple Key Types
Yaron: why are we sending a JWK *as well as* a cert? Are we checking that the cert contains the same public key as the cert?

Justin: I would be OK with the different formats being mutually exclusive somehow. The AS would need to check that they’re the same and throw an error if not. But this would also let the RC send over whatever it’s got for its own keys every time.

jricher:
After discussion with @fimbault , this PR now makes keys single-format-only as previously suggested.

Closes #18

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3428,24 +3428,17 @@ proof of that key MUST be presented in that request. For a key used as part of a
 access token response in {{response-token-single}}, the proof of that key MUST
 be used when presenting the access token.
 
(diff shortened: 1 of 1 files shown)


fimbault (commented):
Not sure about this one, this seems very complex to manage

What's the actual use case for supporting multiple formats ? Unless we have a clear use case, it would seem easier to just say "A key presented by value MUST be a public key in one (and one only) supported format"

If that's really what we want, then the proposed text is OK

fimbault (approved):

PR #444, opened 2022-10-05
make token character set strict

### LINKED ISSUE #24: Access Token Syntax / ABNF
**Yaron**: 3.2.1: The ASCII restriction on "value" is insufficient. Needs to be printable ASCII, 0x20<v<=0x7f. And I'm not sure if this doesn't leave some characters that still need to be escaped.

**Justin**: That’s fair, and it’s currently a SHOULD anyway so I’m fine with making it more restrictive in that case. OAuth 2 currently has the range "%x20-7E”, but I’m not even sure how strictly that’s enforced in the wild.

**Yaron**: JSON is not very liberal here, at least in theory. Quoting RFC 8259: 

> However, the ABNF in this specification allows member names and string values to contain bit sequences that cannot encode Unicode characters. [...] The behavior of software that receives JSON texts containing such values is unpredictable; for example, implementations might return different values for the length of a string value or even suffer fatal runtime exceptions.

jricher:
closes #24

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -52,6 +52,7 @@ normative:
     RFC8174:
     RFC8259:
     RFC8705:
+    RFC9110:
     I-D.ietf-gnap-resource-servers:
     I-D.ietf-httpbis-message-signatures:
     I-D.ietf-httpbis-digest-headers:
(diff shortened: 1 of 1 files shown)


fimbault (commented):
Maybe it's because I don't know what token86 is. Or do we just mean 'token' with a reference to section 5.6.2 of that spec (then that would make sense to me)

fimbault (approved):

PR #446, opened 2022-10-10
update grant definition

### LINKED ISSUE #391: Terminology: "Grant"
The term _Grant (noun)_ is not used as "the act of granting" in this document. It is used as an abstract entity that can be cancelled etc.

So, I suggest to change the definition to:

> Grant
: (verb): to permit an instance of client software to receive some attributes at a specific time and valid for a specific duration and/or to exercise some set of delegated rights to access a protected resource 
(noun): the permission that was given.

fimbault:
closes #391


jricher (approved):
Looks good, I added one suggested punctuation change.

PR #449, opened 2022-10-13
explanation on the use of class_id

### LINKED ISSUE #393: Class_id determined by the AS?
In Sec. 2.3 we say, "The contents and format of this field are up to the AS." Assuming no pre-registration, this doesn't make sense, because the Client has not interacted with the AS yet. And in fact if pre-reg has taken place, we say this field should be ignored!

Maybe say instead: The content of this field is typically statically configured, and MAY be used as a hint to the AS.

fimbault:
closes #393

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1086,6 +1086,8 @@ object with the following fields.
 
 Additional fields are defined in the [Client Instance Fields Registry](#IANA-client-instance).
 
+The "class_id" field can be used in a variety of ways to help the AS make sense of the particular context in which the client is operating. In corporate environments different levels of trust might apply depending on security policies, and might help the AS adjust its own access decisions. It is possible to configure a set of values and rules during a pre-registration, and then later in runtime as a hint to the AS. In other cases, the client runs with a specific AS in mind, so an hardcoded value would acceptable (for instance, a set top box with a "class_id" claiming to be "FooBarTV version 4"). Or yet while the client instance may not have contacted the AS yet, the value of this "class_id" field may also be evaluated by the AS according to a broader context of dynamic use, alongside other related information available elsewhere (for instance, corresponding fields in a certificate). If the AS is not able to interpret the class_id field, it MAY return an "invalid_client" error or choose to return lesser levels of privileges.
+
 The client instance MUST prove possession of any presented key by the `proof` mechanism
 associated with the key in the request. Key proofing methods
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -1086,6 +1086,8 @@ object with the following fields.
...
... AS yet, the value of this "class_id" field may also be evaluated by the AS according to a broader context of dynamic use, alongside other related information available elsewhere (for instance, corresponding fields in AS make sense of the particular context in which the client is operating. In corporate environments different levels of trust might apply depending on security policies, and this field aims to help the AS adjust its own access decisions. It is possible to configure a set of values and rules during a pre-registration, and then later in runtime as a hint to the AS. In other cases, the client runs with a specific AS in mind, so an hardcoded value would acceptable (for instance, a set top box with a "class_id" claiming to be "FooBarTV version 4"). Or while the client instance may not have contacted the AS yet, the value of this "class_id" field may also be evaluated by the AS according to a broader context of dynamic use, alongside other related information available elsewhere (for instance, corresponding fields in a certificate). If the AS is not able to interpret the class_id field, it MAY return an "invalid_client" error or choose to return lesser levels of privileges.
```
jricher, line 1089: I think there are some copy-paste errors here, I tried to trim this down to a single paragraph without the additional sentences.

```suggestion
Both the `display` and `class_id` are self-declarative and thus the AS SHOULD exercise caution in their interpretation, taking them as a hint but not as absolute truth. The `class_id` field can be used in a variety of ways to help the a variety of ways to help the AS make sense of the particular context in which the client instance is operating. In corporate environments, for example, different levels of trust might apply depending on security policies. This field aims to help the AS adjust its own access decisions for different classes of client software. It is possible to configure a set of values and rules during a pre-registration, and then have the client instances provide them later in runtime as a hint to the AS. In other cases, the client runs with a specific AS in mind, so a single hardcoded value would acceptable (for instance, a set top box with a `class_id` claiming to be "FooBarTV version 4"). While the client instance may not have contacted the AS yet, the value of this `class_id` field can be evaluated by the AS according to a broader context of dynamic use, alongside other related information available elsewhere (for instance, corresponding fields in a certificate). If the AS is not able to interpret the class_id field, it SHOULD return an `invalid_client` error or choose to return lesser levels of privileges. See additional discussion of client instance impersonation in {{security-impersonation}}.
```

jricher (changes requested):
This is good content, but needs to be cleaned up. I tried to provide suggested edits to this paragraph here.

jricher (approved):

PR #450, opened 2022-10-13
RS first discovery

### LINKED ISSUE #409: RS-first discovery: security aspects
This discovery method is great for usability but worrying from a security POV. The RS is telling the Client to create a request to an arbitrary AS using an arbitrary access specification and the Client's credentials. The Client doesn't have any way to understand, validate or filter anything in the RS's response, so it cannot protect itself.

By the way, shouldn't we use a semicolon before the `access` parameter?

```
WWW-Authenticate: \
  GNAP as_uri=https://server.example/tx;access=FWWIKYBQ6U56NL1
```

A possible workaround is to add something analogous to a `Referrer` parameter to the AS access request containing the RS address, require the Client to include it, and require that the AS validate that the request is indeed to that RS.

fimbault:
fix typos
closes #409

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4778,11 +4778,11 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
 ## RS-first Method of AS Discovery {#rs-request-without-token}
 
 If the client instance calls an RS without an access token, or with an
-invalid access token, the RS SHOULD be explicit about the fact that GNAP needs to be used to access the resource, by responding with the WWW-Authenticate header field and a GNAP challenge. 
+invalid access token, the RS SHOULD be explicit about the fact that GNAP needs to be used to access the resource, by responding with the WWW-Authenticate header field and a GNAP challenge.
 
 In some situations, the client instance might want to know with which specific AS it needs to negotiate for access to that RS.
 The RS MAY additionally return the address of the GNAP endpoint in the "as_uri" parameter, as well as a referer parameter to indicate which RS initiated the discovery process and an opaque reference that the client instance SHOULD then use in
-its access token request. The "referer" parameter MUST be the URI of the RS, and the client MUST check its value. The opaque reference MUST be sufficient for at least the action the client instance was attempting to take at the RS and MAY be more powerful. 
+its access token request. The "referer" parameter MUST be the URI of the RS, and the client MUST check its value. The opaque reference MUST be sufficient for at least the action the client instance was attempting to take at the RS and MAY be more powerful.
 
 The means for the RS to determine the opaque reference are out of scope of this specification, but some dynamic methods are discussed in
 {{I-D.ietf-gnap-resource-servers}}.
@@ -4828,7 +4828,7 @@ The "Referer" header is a way for the AS to know that the process is initiated t
 If issued, the resulting access token would contain sufficient access to be used at both referenced resources.
 
 Security considerations, especially related to the potential of a [compromised RS]{#security-compromised-rs} redirecting the requests of an otherwise properly authenticated client, need to be carefully considered when allowing such a discovery process.
-This risk can be mitigated by an alternative pre-registration process so that the client knows which AS protects which RS. 
+This risk can be mitigated by an alternative pre-registration process so that the client knows which AS protects which RS.
 
 # Acknowledgements {#Acknowledgements}
 
@@ -6222,8 +6222,8 @@ are supported, if desired. An AS should be generous in supporting many different
 allow different types of client software and client instance deployments.
 
 ## Compromised RS {#security-compromised-rs}
-An attacker may aim to gain access to confidential or sensitive resources. The measures for hardening and monitoring server systems is out of the scope of this document. 
-GNAP generally considers a breach can occur, and therefore advises to prefer key-bound tokens whenever possible, which at least limit the impact of access token leakage by a compromised RS. 
+An attacker may aim to gain access to confidential or sensitive resources. The measures for hardening and monitoring server systems is out of the scope of this document.
+GNAP generally considers a breach can occur, and therefore advises to prefer key-bound tokens whenever possible, which at least limit the impact of access token leakage by a compromised RS.
 
 # Privacy Considerations {#privacy}
 
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4780,9 +4780,9 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
 If the client instance calls an RS without an access token, or with an invalid access token, the RS SHOULD be explicit about the fact that GNAP needs to be used to access the resource, by responding with the WWW-Authenticate header field and a GNAP challenge.
 
 In some situations, the client instance might want to know with which specific AS it needs to negotiate for access to that RS.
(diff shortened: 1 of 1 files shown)


jricher (approved):

fimbault:
Removing the editorial label

draft-ietf-gnap-core-protocol.md
```diff
@@ -4777,42 +4777,31 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
...
-resource reference MUST be sufficient for at least the action
-the client instance was attempting to take at the RS and MAY be more
-powerful.
-The means for the RS to determine the resource reference are out of scope
-of this specification, but some dynamic methods are discussed in
+If the client instance calls an RS without an access token, or with an invalid access token, the RS SHOULD be explicit about the fact that GNAP needs to be used to access the resource, by responding with the WWW-Authenticate header field and a GNAP challenge.
+
+In some situations, the client instance might want to know with which specific AS it needs to negotiate for access to that RS.
+The RS MAY additionally return the address of the GNAP endpoint in the "as_uri" parameter, as well as a referer parameter to indicate which RS initiated the discovery process and an opaque reference that the client instance SHOULD both then use in its access token request. The "referer" parameter MUST be the URI of the RS, and the client MUST check its value to protect itself. The opaque reference MUST be sufficient for at least the action the client instance was attempting to take at the RS and MAY be more powerful.
```
jricher, line 4783:
```suggestion
The RS MAY additionally return the address of the GNAP endpoint in the `as_uri` parameter, a `referrer` parameter to indicate which RS initiated the discovery process, and an opaque `access` reference. The client instance SHOULD then use both the `referrer` and `access` parameters in its access token request. The `referrer` parameter MUST be the URI of the RS, and the client instance MUST check its value to protect itself. The opaque `access` reference MUST be sufficient for at least the action the client instance was attempting to take at the RS and MAY be more powerful.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4777,42 +4777,31 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
...
+
+The means for the RS to determine the opaque reference are out of scope of this specification, but some dynamic methods are discussed in
```
jricher, line 4785:
```suggestion
The means for the RS to determine the value for the `access` reference are out of scope of this specification, but some dynamic methods are discussed in
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4777,42 +4777,31 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
...
-The content of the resource reference is opaque to the client instance.
+
+When receiving the following response from the RS:
 
 ~~~ http-message
 NOTE: '\' line wrapping per RFC 8792
 
 WWW-Authenticate: \
-  GNAP as_uri=https://server.example/tx,access=FWWIKYBQ6U56NL1
+  GNAP as_uri=https://as.example/tx;opaque=FWWIKYBQ6U56NL1;referer=https://rs.example
```
jricher, line 4794:
```suggestion
  GNAP as_uri=https://as.example/tx\
  ;access=FWWIKYBQ6U56NL1\
  ;referer=https://rs.example
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4777,42 +4777,31 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
...
 ~~~
 
-The client instance then makes a request to the "as_uri" as described in
-{{request}}, with the value of "access" as one of the members
-of the `access` array in the `access_token` portion of the request. The
-client instance MAY request additional resources and other information.
-The client instance MAY request multiple access tokens.
+The client instance then makes a request to the "as_uri" as described in {{request}}, with the value of "referer" passed as a header and the "opaque" reference passed unchanged into the `access` array in the `access_token` portion of the request. The client instance MAY request additional resources and other information.
```
jricher, line 4797:
```suggestion
The client instance then makes a request to the `as_uri` as described in {{request}}, with the value of `referrer` passed as an HTTP Referer header field and the `access` reference passed unchanged into the `access` array in the `access_token` portion of the request. The client instance MAY request additional resources and other information.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4777,42 +4777,31 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
...
 
-In this non-normative example, the client instance is requesting a single access
-token using the resource reference `FWWIKYBQ6U56NL1` received from the RS
-in addition to the `dolphin-metadata` resource reference that the client instance
-has been configured with out of band.
+In this non-normative example, the client instance is requesting a single access token using the opaque reference `FWWIKYBQ6U56NL1` received from the RS in addition to the `dolphin-metadata` that the client instance has been configured out of band.
```
jricher, line 4799:
```suggestion
In this non-normative example, the client instance is requesting a single access token using the opaque access reference `FWWIKYBQ6U56NL1` received from the RS in addition to the `dolphin-metadata` that the client instance has been configured with out of band.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4777,42 +4777,31 @@ Additional fields can be defined the [Authorization Server Discovery Fields Regi
...
 
 ~~~ http-message
 POST /tx HTTP/1.1
-Host: server.example.com
+Host: as.example
+Referer: rs.example
```
jricher, line 4804: This header field is a full URL, not just the hostname:

```suggestion
Referer: https://rs.example/resource
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -4829,8 +4818,11 @@ Content-Digest: sha-256=...
 }
 ~~~
 
-If issued, the resulting access token would contain sufficient access to be used
-at both referenced resources.
+The client includes the "Referer" header as a way for the AS to know that the process is initiated through a discovery process.
```
jricher, line 4821:
```suggestion
The client instance includes the Referer header field as a way for the AS to know that the process is initiated through a discovery process at the RS.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -6223,6 +6215,11 @@ instance capable of multiple formats can use [AS discovery](#discovery) to deter
 are supported, if desired. An AS should be generous in supporting many different key formats to
 allow different types of client software and client instance deployments.
 
+## Compromised RS {#security-compromised-rs}
+
+An attacker may aim to gain access to confidential or sensitive resources. The measures for hardening and monitoring server systems is out of the scope of this document.
```
jricher, line 6220:
```suggestion
An attacker may aim to gain access to confidential or sensitive resources. The measures for hardening and monitoring resource server systems (beyond protection with access tokens) is out of the scope of this document, but the use of GNAP to protect a system does not absolve the resource server of following best practices.
```

jricher (changes requested):
I've got a few comments on wording. Mostly though: I'm not sure about changing the `access` field to `opaque`, since it's not just ANY opaque value but specifically an opaque value for use in the access section of the request. As such, I think `opaque` is too generic in this use. If we change the parameter name back I think this is OK.

draft-ietf-gnap-core-protocol.md
fimbault: Ok, we can revert that. My reasoning here was that access might be a bit misleading, as it actually doesn't open or request an access per say, but only forwards the request

draft-ietf-gnap-core-protocol.md
fimbault: yes

PR #451, opened 2022-10-13
dynamic grant endpoint discovery

### LINKED ISSUE #411: Discuss ways to find the grant endpoint
The discovery process defined in GNAP details how to get information about the AS once you know the grant endpoint and can talk directly to it. We do not cover how to get the grant endpoint, except in the one case of it being provided by the RS at runtime in a response. The core document should at least discuss this gap as intentional, and an extension specification (or several) could be written to fill this gap in different circumstances.

fimbault:
closes #411

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4832,6 +4832,10 @@ Content-Digest: sha-256=...
 If issued, the resulting access token would contain sufficient access to be used
 at both referenced resources.
 
+## Dynamic grant endpoint discovery {#grant-discovery}
+
+Except in the previous case where "as_uri" is provided by the RS at runtime, no other method are detailed on how the grant endpoint is provided. This is intentional, as many cases are just fine with static setups (just as in OAuth 2). However, the dynamic nature of GNAP makes it a good candidate for further extensions that would be able to handle more advanced scenarios.
+
 # Acknowledgements {#Acknowledgements}
 
 The editors would like to thank the feedback of the following individuals for their reviews,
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -4832,6 +4832,10 @@ Content-Digest: sha-256=...
 If issued, the resulting access token would contain sufficient access to be used
 at both referenced resources.
 
+## Dynamic grant endpoint discovery {#grant-discovery}
+
+Except in the previous case where "as_uri" is provided by the RS at runtime, no other method are detailed on how the grant endpoint is provided. This is intentional, as many cases are just fine with static setups (just as in OAuth 2). However, the dynamic nature of GNAP makes it a good candidate for further extensions that would be able to handle more advanced scenarios.
+
```
jricher, line 4838: I see what you're going for, but this might be a better way to state it, would like to hear your thoughts:

```suggestion
Additional methods of discovering the appropriate grant endpoint for a given application
are outside the scope of this specification. This limitation is intentional, as many applications
rely on static configuration between the client instance and AS, as is common in OAuth 2.0.
However, the dynamic nature of GNAP makes it a prime candidate for other extensions defining methods
for discovery of the appropriate AS grant endpoint at runtime. Advanced use cases could define
contextual methods for contextually  providing this endpoint to the client instance securely.
Furthermore, GNAP's design intentionally requires the client instance to only know the grant
endpoint and not additional parameters, since other functions and values can be disclosed
and negotiated during the grant process.

```

jricher (changes requested):
I like the idea of having the extra section but I don't think it quite hits what we want to say. I've proposed an alternative text, but think you should take another look at it.

jricher (approved):

PR #457, opened 2022-10-22
rotation grace period

### LINKED ISSUE #456: delay for previous_key
from @adeinega "am I right that previous_key becomes immediately invalid? If it's true, wouldn't be nice to have here some way to rotate them more or less gracefully? It could be useful when there are multiple running instances of the client and it's tough, or even impossible, to update the keys on all these instances immediately (just as an example, the client instance could be deployed as a service in K8S)."

We'll handle this via a (non normative) security consideration

fimbault:
Closes #456

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3535,7 +3535,7 @@ as an extension.
 The security of GNAP relies on the cryptographic security of the keys themselves.
 When symmetric keys are used in GNAP, a key management system or secure key derivation mechanism MUST be used to supply the keys. Symmetric keys MUST NOT be a human memorable password or a value derived from one. Symmetric keys MUST NOT be passed by value from the client instance to the AS.
 
-Additional security considerations apply when [rotating keys] {#security-key-rotation}.
+Additional security considerations apply when [rotating keys]{#security-key-rotation}.
 
 ## Presenting Access Tokens {#use-access-token}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3535,6 +3535,8 @@ as an extension.
 The security of GNAP relies on the cryptographic security of the keys themselves.
 When symmetric keys are used in GNAP, a key management system or secure key derivation mechanism MUST be used to supply the keys. Symmetric keys MUST NOT be a human memorable password or a value derived from one. Symmetric keys MUST NOT be passed by value from the client instance to the AS.
 
+Additional security considerations apply when [rotating keys] {#security-key-rotation}.
```
jricher, line 3538:
```suggestion
Additional security considerations apply when [rotating keys]{#security-key-rotation}.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5890,6 +5892,12 @@ token itself. This approach would make interception of the return from the token
 equivalent to that of a bearer token, since all information required to use the access token
 would be present in the request.
 
+## Key Rotation Policy {#security-key-rotation}
+
+The AS can define its own policy regarding the timeout of the "previous_key", either making it immediately obsolete or allowing for a limited grace period during which both the "previous_key" and the current key can be used for signing requests. Such a grace period can be useful when there are multiple running instances of the client and it's tough, or even impossible, to update the keys on all these instances immediately (just as an example, the client instance could be deployed as a service in kubernetes).
```
jricher, line 5897:
```suggestion
When keys are rotated, there could be a delay in the propagation of that rotation to various components in the AS's ecosystem. The AS can define its own policy regarding the timeout of the `previous_key`, either making it immediately obsolete or allowing for a limited grace period during which both the `previous_key` and the current key can be used for signing requests. Such a grace period can be useful when there are multiple running copies of the client that are coordinated with each other. For example, the client software could be deployed as a cloud service with multiple orchestrated nodes. Each of these copies is deployed using the same key and therefore all the nodes represent the same client instance to the AS. In such cases, it can be difficult, or even impossible, to update the keys on all these copies in the same instant.
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5890,6 +5892,12 @@ token itself. This approach would make interception of the return from the token
...
+
+Narrowly restricting the exposure opportunities for exploit in terms of time, place, and method, especially if the time and place happens only once, makes exploit significantly more difficult. Therefore the "previous_key" (or any previous one before it) is expected to no longer have rotation authority.
```
jricher, line 5899:
```suggestion
The need for accommodating such known delays in the system needs to be balanced with the risk of allowing an old key to still be used. Narrowly restricting the exposure opportunities for exploit at the AS in terms of time, place, and method makes exploit significantly more difficult, especially if the exception happens only once. For example, the AS can reject requests from the `previous_key` (or any previous one before it) to cause rotation to a new key, or at least ensure that the rotation happens to the same known key.

See also the related considerations for token values in {{security-network-management}}.
```

jricher (approved):
Great addition, needs a bit of tweaking in the wording and one cross link, which I've added here. Will apply and merge.

PR #468, opened 2022-11-10
remove previous_key, add rotation requirements

### LINKED ISSUE #470: Remove "previous_key"
The "previous_key" element added in #435 is not necessary in practice because the AS should always be able to look up the current key value based on the access token, and the signature using the previous key proves possession of that key separate from any presentation of the key value or reference.

Closed by #468

jricher:
Closes #470

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3448,21 +3448,22 @@ NOTE: '\' line wrapping per RFC 8792
 
 If the client instance wishes to bind a new presentation key to an access token, the client
 instance MUST present both the new key and previous key in the access token rotation request.
-The client instance makes an HTTP POST as a JSON object with the following fields:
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3830,6 +3821,17 @@ NOTE: '\' line wrapping per RFC 8792
 }
 ~~~
 
+Key proofing methods SHOULD define a mechanism to allow the rotation of keys discussed
+in {{rotate-access-token-key}}. Key rotation mechanisms MUST define a way for presenting
+proof of two keys simultaneously with the following attributes:
+
+- The value of or reference to the new key material MUST be signed by the existing key.
```
fimbault, line 3828: This is a bit redundant with the part in section 6.1.1 (although a bit clearer here)

fimbault (approved):
Makes sense

PR #469, opened 2022-11-10
Add "tag" and other httpsig requirements

### LINKED ISSUE #467: HTTP Signatures Tag element
This was recently added to HTTP Signatures. Please add a standard `tag` value for GNAP messages.

jricher:
Closes #467

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -3893,6 +3893,8 @@ MUST also include the following:
 
 Other message components MAY also be included.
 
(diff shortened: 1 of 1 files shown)


yaronf (approved):
LGTM

fimbault:
From the HTTP message sig spec : "tag: An application-specific tag for the signature as a String value. This value is used by applications to help identify signatures relevant for specific applications or protocols."
Fair enough, but does a MUST requirement improve the security of the protocol ?

yaronf:
> Fair enough, but does a MUST requirement improve the security of the protocol ?

Yes, because it prevents cross-protocol confusion attacks. Where an attacker tries to use a signed HTTP message from another protocol in GNAP, or vice versa.

PR #474, opened 2022-11-24
Update draft from complete readthrough

jricher:
Includes changes through section 3


draft-ietf-gnap-core-protocol.md
```diff
@@ -1481,7 +1496,7 @@ launching a browser on the end user's device, providing a scannable
 image encoding, and printing out a URI to an interactive
 console. While this URI is generally hosted at the AS, the client
 instance can make no assumptions about its contents, composition,
-or relationship to the AS grant URI.
+or relationship to the grant endpoint URI.
```
fimbault, line 1499: Ok but we never explicitly define it. Maybe in section 2 we should just add URI in that sentence : "Sending a request to the grant endpoint" so that it's easier to understand

Also we might want to put a note next to the grant definition saying that in GNAP, there's a URI at the AS to manage this.

draft-ietf-gnap-core-protocol.md
```diff
@@ -2057,9 +2084,7 @@ with a multiple access token structure containing one access token.
 ]
 ~~~
 
-Each access token MAY be bound to different keys with different proofing mechanisms.
-
-The `manage` URI MUST NOT contain the access token `value`.
+The parameters of each access token are separate. For example, each access token MAY be bound to different keys with different proofing mechanisms.
```
fimbault, line 2087: yes, although that example is probably a bit extreme

draft-ietf-gnap-core-protocol.md
```diff
@@ -2834,17 +2897,16 @@ https://server.example.com/tx
...
 
-The "hash_method" value MUST be one of the hash name strings defined in the
+If provided, the "hash_method" value MUST be one of the hash name strings defined in the
 [IANA Named Information Hash Algorithm Registry](https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg).
-
 If the "hash_method" value is not present in the client instance's
 request, the algorithm defaults to "sha-256".
 
-The "sha-256" hash method consists of hashing the input string
+For example, the "sha-256" hash method consists of hashing the input string
```
fimbault, line 2909: I don't understand why we say it's an example. Here we handle sha-256 and just after, we handle sha3 -> suggestion to remove "for example"

draft-ietf-gnap-core-protocol.md
```diff
@@ -3120,20 +3182,24 @@ still needs access to the resource, it will need to start with a new grant reque
...
+request, since the extent and context of the request could have changed.
 
 The client instance MAY include the `access_token` and `subject` fields as described in {{request-token}}
 and {{request-subject}}. Inclusion of these fields override any values in the initial request,
 which MAY trigger additional requirements and policies by the AS. For example, if the client instance is asking for
 more access, the AS could require additional interaction with the RO to gather additional consent.
 If the client instance is asking for more limited access, the AS could determine that sufficient authorization
 has been granted to the client instance and return the more limited access rights immediately.
+If the grant request was previously in the _approved_ state, the AS could decide to remember the larger scale of access rights associated
+with the grant request, allowing the client instance to make subsequence requests of different
```
fimbault, line 3201: typo : subsequence -> subsequent

draft-ietf-gnap-core-protocol.md
```diff
@@ -5748,15 +5841,20 @@ and abused in making needlessly insecure systems.
 
 In GNAP, key-bound access tokens are the default due to their higher security properties. While
 bearer tokens can be used in GNAP, their use should be limited to cases where the simplicity
-benefits outweigh the significant security downsides.
+benefits outweigh the significant security downsides. One common deployment pattern is to use a
+gateway that takes in key-bbound tokens on the outside, and verifies the signatures on the incoming
```
fimbault, line 5845: typo bound

draft-ietf-gnap-core-protocol.md
```diff
@@ -5748,15 +5841,20 @@ and abused in making needlessly insecure systems.
 
 In GNAP, key-bound access tokens are the default due to their higher security properties. While
 bearer tokens can be used in GNAP, their use should be limited to cases where the simplicity
-benefits outweigh the significant security downsides.
+benefits outweigh the significant security downsides. One common deployment pattern is to use a
```
fimbault, line 5844: Maybe (to make it more concrete) we could note that with advanced MFA becoming more mainstream (e.g. FIDO2/webauthn/keypass), the next frontier will be token theft. Therefore having some kind of protection is important

draft-ietf-gnap-core-protocol.md
```diff
@@ -6450,12 +6568,16 @@ to manage at runtime, and systems should generally refuse to fetch a URI if unsu
...
-receivers of these key definitions need to be able to make sure that it's the same
+receivers of these key definitions would need to be able to make sure that it's the same
 key represented in each field and not simply use one of the key formats without checking for
 equivalence. If equivalence were not carefully checked, it is possible for an attacker to insert
 their own key into one of the formats without needing to have control over the other formats. This
 could potentially lead to a situation where one key is used by part of the system (such as
-identifying the client instance) and a different key in a different format in the same message is used for other things, like calculating signature validity.
+identifying the client instance) and a different key in a different format in the same message is
+used for other things (such as calculating signature validity). However, the primary reason to allow
+keys to be presented in multiple formats simultaneously is the assumption that the receiver cannot
```
fimbault, line 6578: Formulation is a bit ambiguous : "primary reason to allow keys to be presented in multiple formats" -> actually not allowed (I understand it's for the sake of explanation, but still)

draft-ietf-gnap-core-protocol.md
```diff
@@ -6512,11 +6634,11 @@ The privacy considerations in this section are modeled after the list of privacy
 
 Surveillance is the observation or monitoring of an individual's communications or activities. Surveillance can be conducted by observers or eavesdroppers at any point along the communications path.
 
-GNAP assumes the TLS protection used throughout the spec is intact. Without the protection of TLS, there are many points throughout the use of GNAP that would lead to possible surveillance.
+GNAP assumes the TLS protection used throughout the spec is intact. Without the protection of TLS, there are many points throughout the use of GNAP that would lead to possible surveillance. Even with the proper use of TLS, surveillance could occur by several parties throughout the protocol.
```
fimbault, line 6637: Ok, but there are potential solutions for this (eg mixnets to prevent traffic analysis by an adversary capable of watching the entire network)

fimbault (commented):
Thanks for the extensive passthrough which are in general looking good, a few comments for clarification

draft-ietf-gnap-core-protocol.md
jricher: Both are actually examples - the definition of the hash function comes from the referenced registry. I'll make them consistent.

draft-ietf-gnap-core-protocol.md
jricher: I was trying to say here that surveillance is still possible in a few ways even if you do tls completely securely, which is what the following sections talk about. It's meant as an introduction, not trying to say that you can break tls to spy on things. Can you think of a better way to get that across?

draft-ietf-gnap-core-protocol.md
jricher: OK, I've added some definitions.

draft-ietf-gnap-core-protocol.md
jricher: added lighter examples

jricher:
@fimbault thanks for the great feedback -- I believe all items have been addressed, please re-review.

fimbault (approved):
Thanks, great work !

PR #475, opened 2022-11-24
change to section 162, verifiable replaced by secure

fimbault:


jricher:
I changed this same line in #474 to just "starts a session with the user"

fimbault:
Ok let's just close this one then !

PR #479, opened 2022-12-05
verify httpsig

### LINKED ISSUE #477: Signature name
In 7.3.1 (use of gnap with HTTP message signatures), we do not mention the signature name (the string for which we always use `sig1`). This is a real issue because multiple signatures are possible on any message. Some options:

* Specify an explicit name, e.g. `sig1` or better yet `gnap`.
* Include a signature name in the `proof` object (but that would preclude the string version of this parameter).

fimbault:
closes #477

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -4150,7 +4150,7 @@ If the HTTP Message includes a message body, the verifier MUST
 calculate and verify the value of the `Content-Digest` header. The verifier MUST validate
 the signature against the expected key of the signer.
 
-A received message MAY include multiple signatures, each with its own label. The verifier MUST examine all included signatures until it finds (at least) one that's acceptable according to its policy.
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -4150,6 +4150,8 @@ If the HTTP Message includes a message body, the verifier MUST
 calculate and verify the value of the `Content-Digest` header. The verifier MUST validate
 the signature against the expected key of the signer.
 
+A received message MAY include multiple signatures, each with its own label. The verifier MUST examine all included signatures until it finds (at least) one that's acceptable according to its policy.
```
jricher, line 4153: Possible strengthening of this statement.

```suggestion
A received message MAY include multiple signatures, each with its own label. The verifier MUST examine all included signatures until it finds (at least) one that's acceptable according to its policy and meets the requirements in this section.
```

fimbault: yes doesn't hurt

PR #494, opened 2023-02-09
User Code Updates

### LINKED ISSUE #481: Unguessability of User Codes
The specification does not require user codes to be unguessable. Section 3.3.3 (Display of a Short User Code) states that user codes have to be unique and should be short-lived, but this does not imply that codes should be unguessable. It seems that Section 13.27 (Exhaustion of Random Value Space) does not apply to user codes, but to values that are clearly random values "such as nonces, tokens, and randomized URIs".

If attackers can guess user codes, the same attack described in https://datatracker.ietf.org/doc/html/rfc8628#section-5.1 is possible.

### LINKED ISSUE #487: Contradiction between the text and the example in section 3.3.3.
This is very minor, but there's a contradiction between the user_code provided in the example in section 3.3.3. and the recommendations that are mentioned in the text above it.

The text says:

`MUST consist of only easily typeable characters (such as letters or numbers)`

And

`It is RECOMMENDED that this code be no more than eight characters in length.`

However, the example provided below it goes against these two recommendations, as it contains a special character (a hyphen) rather than just letters and numbers, and it's longer than eight characters in length:

```
"interact": {
    "user_code": "A1BC-3DFF"
}
```

I'd suggest either tweaking the wording, or tweaking the example so that the user_code provided in the example is aligned to the recommendations mentioned in the text.

jricher:
- clarify that user codes are unguessable, closes #481
- update user code examples, closes #487

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -2189,9 +2189,10 @@ If the client instance indicates that it can
 and the AS supports this mode for the client instance's
 request, the AS responds with a "user_code" field. This field is string
 containing a unique short code that the user
-can type into a web page. This string MUST be
+can type into a web page. To facilitate usability, this string MUST be
 case-insensitive, MUST consist of only easily typeable
-characters (such as letters or numbers). The time in which this
+characters (such as letters or numbers). The string MUST be randomly generated
+so as to be unguessable by an attacker within the time it is accepted. The time in which this
 code will be accepted SHOULD be short lived, such as several
 minutes. It is RECOMMENDED that this code be no more than eight
 characters in length.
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -2690,6 +2692,14 @@ request and MUST be protected by HTTPS or equivalent means.
 When the end user is directed to enter a short code through the ["user_code"](#response-interact-usercode)
 mode, the client instance communicates the user code to the end user and
 directs the end user to enter that code at an associated URI.
+The client instance MAY
```
fimbault, line 2695: Not sure about this memorability thing, or at least I believe it can remain something unspecified (potentially allowed by some implementations if they wish). It's not so easy to make sure this doesn't open new routes of attack. Or else we'd need to explain how to handle the fuzziness.

fimbault (changes requested):
Some comments on the memorability feature

draft-ietf-gnap-core-protocol.md
jricher: I think it's OK here because the fuzziness is only in formatting and only if the resulting format is equivalent to the actual value that needs to be input by the user. This is basically acknowledging something that people already do in practice, but putting the right guides around it in the hopes that things are done correctly.

fimbault: OK

PR #496, opened 2023-02-09
Update interaction extensions

### LINKED ISSUE #484: Requirements on interaction modes defined by extensions
Thinking about possible extensions to GNAP's interaction modes, I had the impression that the specification could be a bit clearer regarding requirements on such extensions.

### Interaction Finish Nonce in Grant Response
From reading Section 3.3, my understanding is that all implementations of GNAP, _including ones using extensions which define additional interaction finish methods_, MUST include a `finish` nonce in the grant response (given that the AS wants to use the finish method offered by the client instance):

https://github.com/ietf-wg-gnap/gnap-core-protocol/blob/052162d7cfae1796f09bc27ddcc8dae85a54402d/draft-ietf-gnap-core-protocol.md?plain=1#L2115-L2116

However, Section 3.3.5 sounds a bit different, only referring to the two interaction finish methods defined by GNAP core:

https://github.com/ietf-wg-gnap/gnap-core-protocol/blob/052162d7cfae1796f09bc27ddcc8dae85a54402d/draft-ietf-gnap-core-protocol.md?plain=1#L2284-L2303

### Interaction Hash
Section 4.2 could be a bit clearer as to what is expected of future extensions to GNAP, it currently says:

https://github.com/ietf-wg-gnap/gnap-core-protocol/blob/052162d7cfae1796f09bc27ddcc8dae85a54402d/draft-ietf-gnap-core-protocol.md?plain=1#L2770-L2772

with the two following sections describing the finish methods defined by GNAP core.
This is, however, somewhat clearer in Section 4.2.3:
>  The AS MUST always provide this hash, and the client instance MUST validate the hash when received.

I.e., Section 4.2.3 implies that AS must (somehow) provide the client instance with an interaction reference and interaction finish nonce. But it may be helpful to make this more explicit throughout the relevant Sections.

### Interaction Start Methods
There seem to be no requirements for future interaction start methods. It may be helpful to document some minimal requirements, e.g., enough information to identify the grant has to be conveyed to AS via RO in the interaction start (such as the redirect URI in GNAP core).

jricher:
Closes #484

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -2779,7 +2779,7 @@ validate the "finish" call.
 All interaction finish methods MUST define a way
 to convey the hash and interaction reference back to the client instance. When an
 interaction finish method is used, the client instance MUST present the interaction
-reference back to the AS as part of its [continuation request](#continuation-after-interaction).
+reference back to the AS as part of its [continuation request](#continue-after-interaction).
 
 Note that in many error cases, such as when the RO has denied
 access, the "finish" method is still enacted by the AS.
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -1630,7 +1630,7 @@ with other values defined by the [Interaction Finish Methods Registry](#IANA-int
 If interaction finishing is supported for this client instance and
 request, the AS will [return a nonce](#response-interact-finish) used by the client
 instance to validate the callback.
-All interaction finish methods MUST use this nonce to allow the client verify the connection
+All interaction finish methods MUST use this nonce to allow the client to verify the connection
 between the pending interaction request and the callback. GNAP does this through the use of the
 interaction hash, defined in {{interaction-hash}}.
 All requests to the callback URI MUST be processed as described in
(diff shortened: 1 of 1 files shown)


jricher:
@SECtim Please take a look at this updated text to see if it addresses your concerns.

draft-ietf-gnap-core-protocol.md
```diff
@@ -1626,12 +1628,19 @@ with other values defined by the [Interaction Finish Methods Registry](#IANA-int
...
 
 If interaction finishing is supported for this client instance and
-request, the AS [returns a nonce](#response-interact-finish) used by the client
+request, the AS will [return a nonce](#response-interact-finish) used by the client
 instance to validate the callback.
-Requests to the callback URI MUST be processed as described in
-{{interaction-finish}}, and the AS MUST require
-presentation of an interaction callback reference as described in
-{{continue-after-interaction}}.
+All interaction finish methods MUST use this nonce to allow the client verify the connection
```
SECtim, line 1633: I think there is a "to" missing in "allow the client TO verify"

SECtim (approved):
Thanks @jricher for taking the time! The proposed text looks good to me (just one possible typo, I'm not a native speaker, so I'm not 100% sure about that).

draft-ietf-gnap-core-protocol.md
jricher:
```suggestion
All interaction finish methods MUST use this nonce to allow the client to verify the connection
```

Issue #493, opened 2023-02-06
Review comments from Yaron


jricher:
Master issue to track comments from @yaronf 

- [x] 1.5: While it is possible to deploy an AS in a stateless environment, such deployments will need a way to manage the current state of the grant request in a secure and deterministic fashion. - I'm not sure such a deployment is possible, e.g. if the state is managed by the client, it would be able to roll back to a previous state and cancel any pending revocations.
- [x] 1.6.1: the generic sequence diagram seems to have a bug, especially when compared to the one in 1.6.4: after message (3), there is interaction with the RO, marked with (B). What then initiates the client's next step, the one marked (4)? Is there polling going on?
- [x] 2.1.1: access token flags - the "bearer" flag is mentioned for requests and responses, but the "durable" flag only for responses. I suggest to add "allowed use" as a new column to the relevant IANA registration registry (Sec. 11.2).
- [x] 2.1.1: "since no flag is provided in this example, is bound to the client instance's key", better be more precise about the flag: since the "bearer" flag is not provided in this example, is bound to the client instance's key.
- [x] 2.2: SAML - please add a reference. Also, is "SAML 2 assertion" well defined? Are there any constraints? If we don't want to give a detailed definition, maybe leave it out.
- [x] 2.3: the example includes both a "jwk" and a "cert" for the public key. I think this contradicts the text in Sec. 7.1: "A key [...] MUST be presented in one and only one supported format".
- [x] 13.12: when we say "natural extension point" I suggest we add a reference to the relevant IANA registry (Sec. 11.5 IMO).
In general it's worth looking at where the word "extension" is mentioned throughout the document. For example, in Sec. 4, "or through some other method defined by an extension of this specification", I don't think we have defined such an extension point, so I would remove that text or add the extension point.
- [x] 2.3: the "display" URI is limited to web addresses, and can simply be a URL.
- [x] 2.3: "SHOULD use a different key for each AS" - I would change it to MUST. We know of one attack, there may be more. If we leave it a “should”, deployers will not have enough information to make an informed decision here.
- [x] 2.3.1: "If the client instance is identified in this manner, the registered key for the client instance MAY be a symmetric key known to the AS." I think this gets it wrong: for symmetric keys, it is the "key" value that's given by reference. The "client" value is unrelated.
[...]

PR #497, opened 2023-02-10
Editorial changes from chair review

jricher:
Addresses #493

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -39,6 +39,7 @@ normative:
           ins: P. Saint-Andre
     RFC2119:
     RFC2397:
+    RFC3339:
     RFC3986:
     RFC4107:
     RFC4648:
@@ -100,6 +101,19 @@ normative:
           ins: B. de Medeiros
         -
           ins: C. Mortimore
+    SAML2:
+        title: "Assertions and Protocol for the OASIS Security Assertion Markup Language (SAML) V2.0"
+        target: https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
+        date: March 2005
+        author:
+          -
+            ins: S. Cantor
+          -
+            ins: J. Kemp
+          -
+            ins: R. Philpott
+          -
+            ins: E. Maler
 
 informative:
     RFC6973:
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -3212,7 +3272,13 @@ If the client instance is asking for more limited access, the AS could determine
 has been granted to the client instance and return the more limited access rights immediately.
 If the grant request was previously in the _approved_ state, the AS could decide to remember the larger scale of access rights associated
 with the grant request, allowing the client instance to make subsequent requests of different
-subsets of granted access. The details of this are out of scope for this specification.
+subsets of granted access. The details of this processing are out of scope for this specification,
+but a one possible approach is as follows:
+
+1. A client instance requests access to A, and is granted by the RO. This results in an access token.
```
fimbault, line 3278: AS (instead of A)

fimbault, line 3278: Need to succinctly explain what A and B are

fimbault (commented):
A minor comment, we can discuss the few remaining sub-issues

draft-ietf-gnap-core-protocol.md
jricher: I was trying to avoid something with built-in semantics, since "write" implies you'd probably also get "read". That's why the example uses the generic "A" and "B" variables for the access rights. We could just as easily use "Foo" and "Bar" here.

draft-ietf-gnap-core-protocol.md
```diff
@@ -3212,7 +3272,13 @@ If the client instance is asking for more limited access, the AS could determine
...
+2. The client instance later modifies the grant request to include A+B. Since the client instance was previously granted A under this grant request, the RO is prompted to allow the client instance access to A+B. This results in a new access token.
+3. The client instance makes another grant modification to ask only for B. Since the client instance was previously granted A+B under this grant request, the RO is not prompted and the access to B is granted in a new access token, with no access to A.
+3. The client instance makes another grant modification to ask only for A. Since the client instance was previously granted A+B under this grant request, the RO is not prompted and the access to A is granted in a new access token, with no access to B.
```
jricher, line 3281:
```suggestion
1. A client instance requests access to `Foo`, and is granted by the RO. This results in an access token, `AT1`.
2. The client instance later modifies the grant request to include `Foo` and `Bar` together. Since the client instance was previously granted `Foo` under this grant request, the RO is prompted to allow the client instance access to `Foo` and `Bar` together. This results in a new access token, `AT2` This access token has access to both `Foo` and `Bar`. The rights of the original access token `AT1` are not modified.
3. The client instance makes another grant modification to ask only for `Bar`. Since the client instance was previously granted `Foo` and `Bar` together under this grant request, the RO is not prompted and the access to `Bar` is granted in a new access token, `AT3`. This new access token does not allow access to `Foo`.
4. The original access token `AT1` expires and the client seeks a new access token to replace it. The client instance makes another grant modification to ask only for `Foo`. Since the client instance was previously granted `Foo` and `Bar` together under this grant request, the RO is not prompted and the access to `Foo` is granted in a new access token, `AT4`. This new access token does not allow access to `Bar`.

All four access tokens are independent of each other and associated with the same underlying grant request. Each of these access tokens could possibly also be rotated using token management, if available. For example, instead of asking for a new token to replace `AT1`, the client instance could ask for a refresh of `AT1` using the rotation method of token management. This would result in a refreshed `AT1`, potentially with a different token value and expiration from the original `AT1` but with the same access rights of allowing only access to `Foo`.
```

fimbault: Great

PR #501, opened 2023-02-21
details on refresh tokens

### LINKED ISSUE #485: Support for Refresh Tokens?
In section 6.1. of the document where it discusses Access Token Rotation, I'm slightly surprised by the lack of support for Refresh Tokens? From my interpretation of the specification, the only valid way to rotate an expired Access Token is to pass in this same **expired** Token (signing the request with the appropriate cryptographic key), and then the AS expected to validate this expired Access Token and conclude that it is valid? I also can't see any provision within the specification for the AS to (optionally) issue a Refresh Token along with the Access Token(s) that it sends to Client?

There is something that feels quite wrong about expecting the AS to determine that an expired Access Token has successfully passed validation. On top of this, the lack of Refresh Token support could be confusing for someone who is familiar with using OAuth 2.0 - at the very least, I think Refresh Tokens should be mentioned in the document with a short explanation as to why they aren't supported (if this is indeed the case, and if this is a deliberate decision). I know the aim here is to have a fresh start and to not automatically replicate all of the concepts and details from OAuth, but I wasn't aware that Refresh Tokens was one of those concepts that needed removing.

I'd welcome your thoughts on these points.

fimbault:
closes #485

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -697,7 +697,7 @@ An example set of protocol messages for this method can be found in {{example-no
 
 In this example flow, the client instance receives an access token to access a resource server through
 some valid GNAP process. The client instance uses that token at the RS for some time, but eventually
-the access token expires. The client instance then gets a new access token by rotating the
+the access token expires. The client instance then gets a refreshed access token by rotating the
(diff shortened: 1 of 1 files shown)


jricher (approved):
I like this approach - I think a light touch is the right way to go about it.

PR #505, opened 2023-05-03
Add access control to token management API

### LINKED ISSUE #493: Review comments from Yaron

### LINKED ISSUE #502: 6.1: Requiring an access token to be unexpired when it is rotated would simplify the text as well as the implementation logic (e.g. around expiration vs. revocation). Why not have the client responsible to rotate the token in a timely manner? If it doesn't, it can always request a new one.

jricher:
Adds the token management token to the token management API calls for rotating and revoking existing access tokens.

Closes #502
Closes #493

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -716,7 +716,7 @@ An example set of protocol messages for this method can be found in {{example-no
 In this example flow, the client instance receives an access token to access a resource server through
 some valid GNAP process. The client instance uses that token at the RS for some time, but eventually
 the access token expires. The client instance then gets a refreshed access token by rotating the
-expired access token at the AS using the token's management URI.
+expired access token's value at the AS using the token management API.
 
 ~~~ aasvg
 {::include diagram/refresh.md}
@@ -1824,14 +1824,16 @@ In this example, the AS is returning a bearer [access token](#response-token-sin
 an opaque identifier.
 
 ~~~ json
-NOTE: '\' line wrapping per RFC 8792
-
 {
     "access_token": {
         "value": "OS9M2PMHKUR64TB8N6BW7OZB8CDFONP219RP1LT0",
         "flags": ["bearer"],
-        "manage": "https://server.example.com/token/PRY5NM33O\
-            M4TB8N6BW7OZB8CDFONP219RP1L",
+        "manage": {
+            "uri": "https://server.example.com/token/PRY5NM33O",
+            "access_token": {
+                "value": "B8CDFONP21-4TB8N6.BW7ONM"
+            }
+        }
     },
     "subject": {
         "sub_ids": [ {
@@ -1895,6 +1897,7 @@ contains a JSON object with the following properties.
     client instance's key used in the request and MUST NOT be a bearer token. As a consequence,
     the `flags` array of this access token MUST NOT contain the string `bearer` and the
     `key` field MUST be omitted.
+    This access token MUST NOT have a `manage` field.
     The client instance MUST present the continuation access token in all requests to the continuation URI as described in {{use-access-token}}.
     REQUIRED.
 
@@ -1951,16 +1954,14 @@ properties.
     [token request](#request-token), if present.
     REQUIRED for multiple access tokens or if a `label` was included in the single access token request, OPTIONAL for a single access token where no `label` was included in the request.
 
-`manage` (string):
-: The management URI for this
-    access token. This URI MUST be an absolute URI.
+`manage` (object):
+: Access information for the token management API for this access token.
+    The management URI for this
+    access token.
     If provided, the client instance MAY manage its access
-    token as described in {{token-management}}. This management
-    URI is a function of the AS and is separate from the RS
+    token as described in {{token-management}}.
+    This management API is a function of the AS and is separate from the RS
     the client instance is requesting access to.
-    This URI MUST NOT include the
-    access token value and SHOULD be different for each access
-    token issued in a request.
     OPTIONAL.
 
 `access` (array of objects/strings):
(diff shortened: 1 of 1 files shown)


fimbault (approved):
This PR is fine

fimbault (approved):
This PR is fine

fimbault (approved):
This PR is fine

fimbault (approved):
This PR is fine

fimbault (approved):
This PR is fine

fimbault (approved):
This PR is fine

PR #506, opened 2023-05-05
key rotation update

jricher:
Update how key rotation happens with HTTP Signatures

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -6899,6 +6899,9 @@ Throughout many parts of GNAP, the parties pass shared references between each o
 
 > Note: To be removed by RFC editor before publication.
 
+- -14
+    - Fix key rotation with HTTP Signatures based on security analysis.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -6867,6 +6899,9 @@ Throughout many parts of GNAP, the parties pass shared references between each o
 
 > Note: To be removed by RFC editor before publication.
 
+- -14
+    - Fix key rotation with HTTP Signatures based on security analysis.
```
fimbault, line 6903: Which security analysis are we referring to ?

fimbault (approved):
Makes sense

draft-ietf-gnap-core-protocol.md
jricher: It was done on HTTPSig, thread starts here: https://lists.w3.org/Archives/Public/ietf-http-wg/2023JanMar/0062.html

PR #509, opened 2023-06-26
User Codes

jricher:
- clarify user code character treatment
- changelog


yaronf:
@jricher You were too quick to merge... The reason for ASCII is not ease of use (it's very easy to type Hebrew characters on my keyboard), rather it is to simplify character normalization and ensure that "case insensitive" is well defined.

PR #513, opened 2023-10-03
AD Review Text Updates

jricher:
Update text based on AD review of core specification:

https://mailarchive.ietf.org/arch/msg/txauth/9Z-nHJsHEDhREiXFl1hDatizb7Q/

### DIFF
--- a/draft-ietf-gnap-core-protocol.md
+++ b/draft-ietf-gnap-core-protocol.md
@@ -321,7 +321,7 @@ Looking back on each trust relationship:
 
 - end user/RO: this relationship exists only when the end user and the RO are different, in which case the end user needs some out of band mechanism of getting the RO consent (see {{authorization}}). GNAP generally assumes that humans can be authenticated thanks to identity protocols (for instance, through an id_token assertion in {{request-subject}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-gnap-core-protocol.md
```diff
@@ -1431,7 +1431,8 @@ its capabilities and what is allowed to fulfill the request.
 
 In this non-normative example, the client instance is indicating that it can [redirect](#request-interact-redirect)
 the end user to an arbitrary URI and can receive a [redirect](#request-interact-callback-redirect) through
-a browser request.
+a browser request. Note that the client instance does not accept a push-style callback. This pattern
+is common for web-based client software.
```
fimbault, line 1435: I would clarify "This redirect pattern" (to make sure we don't misunderstand with the note on push style callback that comes just before)

fimbault (approved):
This all makes sense and is aligned with the explanations on the mailing list

draft-ietf-gnap-core-protocol.md
jricher, line 1435:
```suggestion
a browser request. Note that the client instance does not accept a push-style callback.
The pattern of using a redirect for both interaction start and finish is common for web-based client software.
```

PR #514, opened 2023-10-18
add section on as-provided keys

jricher:


jricher:
@yaronf good call, we forgot to put in a proper forward reference -- I'll go add that

PR #533, opened 2024-03-08
Add IANA registration for JWS types

jricher:


OR13 (approved):

draft-ietf-gnap-core-protocol.md
```diff
@@ -5458,6 +5472,184 @@ This specification requests registration of the following scheme in the
...
+* Restrictions on usage: none
+* Author: IETF GNAP Working Group, txauth@ietf.org
+* Change Controller: IETF
+* Provisional registration?  No
+
+
+To indicate that the content is a GNAP message to be bound with a detached JWS mechanism:
+
+* Type name: application
+* Subtype name: gnap-binding+jwsd
```
OR13, line 5532: you don't have examples using this anymore, will it be clear how it is used?

jricher: Yes we do: https://deploy-preview-533--gnap-core-protocol-editors-draft.netlify.app/draft-ietf-gnap-core-protocol.html#name-detached-jws

draft-ietf-gnap-core-protocol.md
```diff
@@ -5458,6 +5472,184 @@ This specification requests registration of the following scheme in the
...
+* Encoding considerations: binary; JWS values are a series of base64url-encoded values (some of which may be the empty string) separated by period ('.') characters.
+* Interoperability considerations: n/a
+* Fragment identifier considerations: n/a
+* Security considerations: See {{security}} of {{&SELF}}
+* Author: IETF GNAP Working Group, txauth@ietf.org
+* Author/Change controller: IETF
+
+To indicate the content is a compact form JWS as defined by {{RFC7515}} with detached payload:
+
+* Name: Detached JWS
```
selfissued, line 5642: `jwsd` is unnecessary and should be deleted.  Per https://www.rfc-editor.org/rfc/rfc7515.html#appendix-F, you can tell whether the payload is detached or not by whether the payload field is empty or not.

draft-ietf-gnap-core-protocol.md
```diff
@@ -5458,6 +5472,184 @@ This specification requests registration of the following scheme in the
...
+the "Structured Syntax Suffix" registry {{IANA.StructuredSuffix}} in
+the manner described in {{RFC6838}}, which can be used to indicate that
+the media type is encoded as a compact form JWS with either attached or detached payload content.
+
+To indicate the content is a compact form JWS as defined by {{RFC7515}} with attached payload:
+
+* Name: JWS
+* +suffix: +jws
+* References: {{&SELF}}
+* Encoding considerations: binary; JWS values are a series of base64url-encoded values (some of which may be the empty string) separated by period ('.') characters.
```
selfissued, line 5633:
```suggestion
* Encoding considerations: binary; JWS values using the compact serialization are a series of base64url-encoded values (some of which may be the empty string) separated by period ('.') characters.
```

draft-ietf-gnap-core-protocol.md
selfissued:
```suggestion
* Subtype name: gnap-binding+jws
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5458,6 +5472,184 @@ This specification requests registration of the following scheme in the
...
+* Intended usage: COMMON
+* Restrictions on usage: none
+* Author: IETF GNAP Working Group, txauth@ietf.org
+* Change Controller: IETF
+* Provisional registration?  No
+
+To indicate that the content is a GNAP token rotation message to be bound with a detached JWS mechanism:
+
+* Type name: application
+* Subtype name: gnap-binding-rotation+jwsd
```
selfissued, line 5578:
```suggestion
* Subtype name: gnap-binding-rotation+jws
```

draft-ietf-gnap-core-protocol.md
```diff
@@ -5458,6 +5472,184 @@ This specification requests registration of the following scheme in the
...
+* Person & email address to contact for further information: IETF GNAP Working Group, txauth@ietf.org
+* Intended usage: COMMON
+* Restrictions on usage: none
+* Author: IETF GNAP Working Group, txauth@ietf.org
+* Change Controller: IETF
+* Provisional registration?  No
+
+##  Structured Syntax Suffix Registration
+
+This section requests registration of the "+jws" and "+jwsd" structured syntax suffixes in
```
selfissued, line 5623:
```suggestion
This section requests registration of the "+jws" structured syntax suffix in
```

selfissued (changes requested):
`jwsd` is unnecessary and should be deleted.  Per https://www.rfc-editor.org/rfc/rfc7515.html#appendix-F, you can tell whether the payload is detached by whether the payload field is empty.

jricher:
@selfissued I'm fine with changing the values here, but the question is more what the appropriate containment is at the protocol level. I don't believe the signaling you are referring to works in this case. There are two JOSE-based mechanisms defined by GNAP, one uses JWS as the message payload, effectively replacing the JSON with its own type -- that's the "Attached JWS" payload. The other keeps the payload as-is (JSON) but uses JWS to create a detached signature. The payload of the JWS then becomes a hash of the content instead of the content itself. Within GNAP these are signaled by separate field values, but shouldn't they be separate `typ` values as well, as per the best practice RFCs? 

If we remove `+jwsd` as a suffix, would you then recommend that the distinction move to the root types instead? If we go that route, I think we would probably just remove the subtypes entirely. So instead of `gnap-binding+jwsd` we'd have `gnap-binding-jwsd`, instead of `gnap-binding+jws` we'd have `gnap-binding-jws`. Would that address your concerns?

selfissued:
Your suggestion above not using structured suffixes would be fine.  Or use something like `gnap-normal-binding+jws` and `gnap-detached-binding+jws` (feel free to pick more appropriate names), if you need to make the distinction at the Content-Type level, rather than after you have the content, if you still want the media type to indicate that the content is a JWS using the Compact Serialization.  Your call.

The core of my review is to not create a separate `jwsd` media type and structured suffix.

Oh, and to be clear in the registration text that `+jws` is about the Compact Serialization.

draft-ietf-gnap-core-protocol.md
```diff
@@ -142,6 +144,18 @@ informative:
...
+      author:
+        -
+          fullname: IANA
+      title: Media Types
+    IANA.StructuredSuffix:
+      target: "https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml"
+      author:
+        -
+          fullname: IANA
+      title: Structured Syntax Suffixs
```
jricher, line 158:
```suggestion
```

selfissued (approved):

PR #535, opened 2024-03-08
SEC AD Review

jricher:
- add invalid request reference
- from normative to declarative
- add requirement for class processing
- make pre-registration declarative
- change reject/request to process and add an example
- clarify activity on unknown keys and symmetric keys.
- clarify example introductions
- fix multi-signature example
- clarify error codes
- fix errors


OR13 (approved):

paulwouters (approved):
Thanks for the quick turnaround!

PR #539, opened 2024-08-26
Nit: multiple components, but client instance not plural

michielbdejong:
This sentence is about one client instance (not plural) consisting of several components that communicate between them.

Although writing 'client instances' (plural) here is not incorrect, is also doesn't seem necessary, and it feels a bit confusing, possibly suggesting that communication between client instances would also take place.


michielbdejong:

PR #542, opened 2024-08-27
Typo: would be acceptable

michielbdejong:
Somewhere halfway that long line:

> runs with a specific AS in mind, so a single hardcoded value would _**be**_ acceptable (for instance, a set top box with a `class_id` claiming to


michielbdejong:

PR #543, opened 2024-08-27
Give context to JSON block of request continuation response

michielbdejong:
Most other examples are preceded by a short sentence that say "In the following non-normative example, ...", but it seems for this one that introductory sentence is missing.


michielbdejong:

PR #544, opened 2024-08-27
Fix management token description copy-paste error

michielbdejong:
It seems this paragraph about the management token was copy-pasted from the description of the request continuation token and not yet fully adapted. This commit attempts to fix that.

Adding the 'itself' word doesn't change the meaning but I thought it makes the statement "the management token should not have a management token" a little bit more readable by acknowledging the self-reference the reader will likely remark while reading that.


michielbdejong:

